Issue #22, opened 2021-06-08
Add a prop to control  show/hide disabled hours, or minutes, seconds


angela-1:
## What problem does the feature solve?

When I use `time-picker` to quickly select a time, I wish to just leave the value not disabled on the panel.

Such as I disable every 10 minutes(enable 0, 10, 20...), to choose 08:00, 08:10, 08:20... 
But the panel show 9 disabled value and 1 enable value.
It not make me choose value faster but slow down.

So i think if you can add a prop to control the show/hide of disabled hours(minutes, seconds), because these values
have been `disabled` by user, means user don't want to use them, so these disabled value should not be shown by default.

## What does the proposed API look like?

The disabled value should not show by default.

```vue
<template>
  <n-time-picker
    v-model:value="time0"
    :show-disabled="true"
    :is-hour-disabled="isHourDisabled"
    :is-minute-disabled="isMinuteDisabled"
    :is-second-disabled="isSecondDisabled"
  />
</template>
```

|name|type|default value|
|  ----  | ----  |----  |
|show-disabled|boolean|false|

Thank your!

07akioni:
Add `hourStep` `minuteStep` and `secondStep` prop can solve the problem. Are you willing to make a PR? It won't be hard.

angela-1:
The `xxStep` can be done by `isMinuteDisabled` function like this:

```js
function isMinuteDisabled(minute, selectedHour) {
  if (selectedHour === null) return false;
  return minute % 10 === 0 ? false : true;
}
```
and add one `show-disabled` prop can get goal. 
It is clear that `is-xx-disabled` controls disabled values and `show-disabled` controls show/hide.


Your suggestions has a question.
How to deal with the relationship between `xxStep` and `is-xx-disabled`. If a user set both `xxStep` 
and `is-xx-disabled`, there are two options: 
1. which one (`step` `disabled`) has higher priority? 
2. user should disabled all the values or just disabled values after filtered by `step`?

I think it is too heavy to add 3 prop for this and will puzzle user, "should i set step or set disabled value? It's not clear"

So far, I don't have a pr this feature: add `show-disabled` prop. Sorry!
After this, I will have a try.

Thanks for reply! :smile:

07akioni:
> 2\. user should disabled all the values or just disabled values after filtered by `step`?

I thinks disabled can only controll the what does the item looks like (and whether it can be click). `step` and `disabled` shouldn't affect each other.

[...]

PR #29, opened 2021-06-08
feat:add showDisable prop to time-picker

w1301625107:
#22

给 time-picker 添加 showDisable 属性


07akioni:
I'm not sure whether `show-disabled` is a good choice.

PR #109, opened 2021-06-14
feat: progress and button-group supports wai-aria

lzj822:


07akioni:
image
As you can see in chrome devtools, there are some generic DOM which shouldn't appear in a11y tree. You need to add `role="none"` for those DOM node.

07akioni:
Button group is nice.

07akioni (commented):
1. Line generic DOM should be fixed.
image

2. `aria-hidden` for svg may be better.
image

3. `aria-value` is better to be set on `.n-progress` but not `progress-fill`. Then make `.n-progress-content` `role="none"` and `.n-progress-graph` `aria-hidden`.
image

4. Please check a11y tree in devtools and make sure at least for there's no generic DOM & SVG dom in `line` and `circle` typed progress. (Also don't forget to make text content displayed in a11y tree).

lzj822:
当 indicatorPlacement === 'inside' 时，无论 showIndicator 为何值，都将会显示indicator，不知道有什么特殊的考虑在其中，我认为这是一个bug。 这个问题导致了 不能用 showIndicator 来判断 role  为progressbar 还是 none。
如果可以，将在下次commit中将 原有判断逻辑 indicatorPlacement === 'inside' 改为 showIndicator && indicatorPlacement === 'inside'

07akioni:
1. For circle progress, there is still a useless generic dom which should be set as `role="none"`.
image

2. In the most basic progress bar, the role shouldn't be `"none"` but `"progressbar"`.
image

3. For a types progress bar & circle, the role shouldn't be `"none"` but `"progressbar"`.
image
image

4. Useless generic dom in the following example
image

PR #132, opened 2021-06-15
Fix docs: on n-layout-footer not have contentStyle attribute

WalkAlone0325:


07akioni:
Please checkout from docs branch. Now commits in main branch is mixed.

PR #160, opened 2021-06-16
feat(tree): support `disabled` & `checkboxDisabled` on option. fix cl…

07akioni:
…ick on indent won't trigger select in block line mode


07akioni:
ref #155

PR #161, opened 2021-06-16
feat: InputNumber support keyboard events ArrowUp and ArrowDown opera…

amadeus711:
…tions

### DIFF
--- a/src/input-number/src/InputNumber.tsx
+++ b/src/input-number/src/InputNumber.tsx
@@ -293,14 +293,12 @@ export default defineComponent({
         if (value !== false) {
           inputInstRef.value?.deactivate()
         }
-      }
-      if (e.code === 'ArrowUp') {
+      } else if (e.code === 'ArrowUp') {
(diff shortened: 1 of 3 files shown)


src/input-number/src/InputNumber.tsx
```diff
@@ -294,6 +294,18 @@ export default defineComponent({
           inputInstRef.value?.deactivate()
         }
       }
+      if (e.code === 'ArrowUp') {
+        const value = deriveValueFromDisplayedValue()
+        if (value !== false) {
+          doAdd()
+        }
+      }
+      if (e.code === 'ArrowDown') {
```
07akioni, line 303: else if

src/input-number/src/InputNumber.tsx
```diff
@@ -294,6 +294,18 @@ export default defineComponent({
           inputInstRef.value?.deactivate()
         }
       }
+      if (e.code === 'ArrowUp') {
```
07akioni, line 297: else if

PR #177, opened 2021-06-17
feat:n-input Support hidden password

doom-9:


07akioni:
It seems duplicate with https://github.com/TuSimple/naive-ui/pull/179

We can wait for that PR and don't modify the PR for a while.

There's no need for a `NBaseEye` since it won't be reuse.

PR #242, opened 2021-06-22
fix(n-input-number): clear input box but value not null

kungege:


07akioni:
Fixed in https://github.com/TuSimple/naive-ui/pull/259

Issue #251, opened 2021-06-22
input-number doesn't change with  /- buttons if you delete its content at least once


wc-matteo:
### TuSimple/naive-ui version (版本)
2.13.0

### Vue version (Vue 版本)
3.1.2

### Browser and its version (浏览器及其版本)
Chrome 91.0.4472.106

### System and its version (系统及其版本)
Windows 10

### Reappearance link (重现链接)
https://codesandbox.io/s/0gse4

### Reappearance steps (重现步骤)
Press   in the min field (notice it increments)
Delete the value from the field (backspace)
Press   again

### Expected results (期望的结果)
The value increments like the first time

### Actual results (实际的结果)
The value doesn't increment. Nothing happens

PR #252, opened 2021-06-23
fix(input-number): doesn't change with /- buttons if you delete its c…

doom-9:
…ontent at least once(#251)


07akioni:
The scene is rather complex. The current solution is not enough.

Sorry for not saying I'm doing the job.

07akioni:
Fixed in #259

PR #269, opened 2021-06-23
ci(dependabot): update github-action when needed

loctran016:


07akioni (commented):
Is all modification essential? Also I found indent is changed.

loctran016:
This isn't must-have PR, but may be nice to have. I guess the indent is changed because I use `tab` instead of `spacebar`.

07akioni:
> This isn't must-have PR, but may be nice to have. I guess the indent is changed because I use `tab` instead of `spacebar`.

I'm not sure if periodically update github action worth it (We will spend time handling those PR and the more important thing is you can't test whether the updated action is valid before it's merged). You can see ant-design is still using many old version actions and it works fine.

Issue #288, opened 2021-06-24
clicking on the date-picker/time-picker icon doesn't open the menu


wc-matteo:
### TuSimple/naive-ui version (版本)
2.14.0

### Vue version (Vue 版本)
3.1.2

### Browser and its version (浏览器及其版本)
Chrome 91.0.4472.114

### System and its version (系统及其版本)
Windows 10

### Reappearance link (重现链接)
https://codesandbox.io/s/3hfmi

### Reappearance steps (重现步骤)
Click on the icon inside the input

### Expected results (期望的结果)
The selector opens

### Actual results (实际的结果)
The selector doesn't open

### Remarks (补充说明)
It happens every time for the time-picker and when there's `clearable` for the date-picker

PR #292, opened 2021-06-24
Fix: Input clear button placeholder prevent clicking on actual component [#288]

delowardev:
fix https://github.com/TuSimple/naive-ui/issues/288


src/_internal/clear/src/Clear.tsx
```diff
@@ -31,7 +31,13 @@ export default defineComponent({
   render () {
     const { clsPrefix } = this
     return (
-      <div class={`${clsPrefix}-base-clear`} data-clear>
+      <div
+        class={[
+          `${clsPrefix}-base-clear`,
+          this.show ? '' : `${clsPrefix}-base-clear--no-pointer`
+        ]}
```
07akioni, line 38: This is not the desired way, you may move `data-clear` to `.n-base-clear__clear`.

Since in date-picker we determine if clear is clicked by this.
image

src/input/src/styles/input.cssr.ts
```diff
@@ -305,6 +305,9 @@ export default c([
       justify-content: center;
       color: var(--suffix-text-color);
     `, [
+      cB('base-clear--no-pointer', {
```
Volankey, line 308: I think you should not write the code in `n-input`, because in some cases user may want the icon to respond to click events.

07akioni: The code should be removed.

delowardev: https://github.com/TuSimple/naive-ui/pull/292/files

now should be ok, thanks

PR #339, opened 2021-06-30
feat(carousel): add trigger prop

kungege:


07akioni (commented):
Please use mouseenter to trigger hover, mouseover is performance consuming.

PR #365, opened 2021-07-03
feat(title attribute): Add native title attributes where necessary

yingpengsha:
## Meaning

In cases where text is likely to be truncated and ignored, you need a way to see the ignored text.
It's can improves usability for impaired visitors using assisted devices.

## Results

| Component | Screenshots |
| --------   | ----- |
| Selection | image |
|  Badge | image |
|  Calendar | image |
|  Transfer | image |
|  Anchor | image |

## Refer

- [HTML 5.2: 3. Semantics, structure, and APIs of HTML documents](https://www.w3.org/TR/html52/dom.html#the-title-attribute)
- [title - HTML: HyperText Markup Language | MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title)

### DIFF
--- a/CHANGELOG.en-US.md
+++ b/CHANGELOG.en-US.md
@@ -6,6 +6,7 @@
 
 - `n-loading-bar` export `LoadingBarApi` type
 - `n-image` add `imgProps` prop
+- Add native `title` attributes to some components to enhance the experience
 
 ## 2.15.2 (2021-07-02)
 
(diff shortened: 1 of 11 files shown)


yingpengsha:
@07akioni PTAL

yingpengsha:
Hold on, There is a commit on the way.

yingpengsha:
Done

PR #429, opened 2021-07-07
fix(n-result): result svg cannot show

Talljack:


Volankey (approved):

07akioni:
fix: result svg cannot show (#429)

=>

fix(result): result svg cannot show (#429)

@Talljack

PR #576, opened 2021-07-22
fix(data-table): cover container width by table content width

bljessica:


Volankey:
Same as https://github.com/TuSimple/naive-ui/pull/575.

PR #614, opened 2021-07-23
Export light theme object

djmattyg007:
This matches the existing `darkTheme` object that's already exported.

I need this because I want to pass theme objects around as props,
and I don't want to make that prop nullable for type safety reasons.

### DIFF
--- a/CHANGELOG.en-US.md
+++ b/CHANGELOG.en-US.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## Pending
+
+### Feats
+
+- Export `lightTheme` theme object (to match `darkTheme`)
+
 ## 2.15.6 (2021-07-23)
 
 ### Feats
--- a/CHANGELOG.zh-CN.md
+++ b/CHANGELOG.zh-CN.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## Pending
+
+### Feats
+
+- Export `lightTheme` theme object (to match `darkTheme`)
+
 ## 2.15.6 (2021-07-23)
 
 ### Feats
--- a/src/index.ts
+++ b/src/index.ts
@@ -10,8 +10,8 @@ export * from './composables'
 
 // component themes
 export * from './styles'
-// composed global theme, createTheme from component themes util
-export { darkTheme, createTheme } from './themes'
+// composed global themes, createTheme from component themes util
+export { lightTheme, darkTheme, createTheme } from './themes'
 
 export { c, cE, cM, cB, cNotM } from './_utils/cssr'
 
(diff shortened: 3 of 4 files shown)


CHANGELOG.zh-CN.md
```diff
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## Pending
+
+### Feats
+
+- Export `lightTheme` theme object (to match `darkTheme`)
```
Talljack, line 7:
```suggestion

- 导出`lightTheme` 主题对象 (对应 `darkTheme`)
```

CHANGELOG.en-US.md
```diff
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## Pending
+
+### Feats
+
+- Export `lightTheme` theme object (to match `darkTheme`)
```
Talljack, line 7:
```suggestion
- Export `lightTheme` theme object (to match `darkTheme`).
```

07akioni:
Nullable is safe for ConfigProvider. I don't think exporting it is necessary.

djmattyg007:
Would you be able to help me understand why it makes sense for `darkTheme` to be exported, but not `lightTheme`?

07akioni:
> Would you be able to help me understand why it makes sense for `darkTheme` to be exported, but not `lightTheme`?

All the component has a builtin default theme.

Long time ago I'm think of whether to give component a default theme, after some time I decided to add default theme object to every component.

I don't want user to import two object if they only need 1 component.

djmattyg007:
The problem with your reasoning is that I can still import the required object like so:

```
import { lightTheme } from "naive-ui/es/themes/light";
```

All this PR does is make peoples' lives a bit simpler.

I also don't think it's a good idea to pre-judge all the possible use cases someone may have for wanting to import this object. Having direct access to it is the easiest way to be able to manipulate it however a developer sees fit, and they may want to do things no one has considered before.

djmattyg007:
Another reason why someone might want direct access to the theme object (rather than relying on `null` to do the right thing) is for type safety. Right now, I have a computed ref that looks like so:

```
    const actualTheme = computed(() => {
      if (actualThemeName.value === "light") {
        return null;
      } else if (actualThemeName.value === "dark") {
        return darkTheme;
      } else {
        throw new Error("Unrecognised theme selected.");
      }
    });
```

This unfortunately means the value is nullable, which makes it more awkward to deal with. If I could pass the `lightTheme` object around, this value would no longer be nullable.

07akioni:
> Another reason why someone might want direct access to the theme object (rather than relying on `null` to do the right thing) is for type safety. Right now, I have a computed ref that looks like so:

Just view null as light theme, this won't do harm.

Issue #618, opened 2021-07-23
NTreeSelect: prop checkable doesn't work


shuaibishun:
### TuSimple/naive-ui version (版本)
2.15.6

### Vue version (Vue 版本)
3.2.0-beta.4

### Browser and its version (浏览器及其版本)
Chrome

### System and its version (系统及其版本)
MacOS

### Reappearance link (重现链接)
https://codesandbox.io/s/3qy1v

### Reappearance steps (重现步骤)
prop checkable doesn't work

### Expected results (期望的结果)
fix

### Actual results (实际的结果)
fix

amadeus711:
Sorry, this is NTreeSelect's demo bug, you can refer to https://codesandbox.io/s/14318

PR #619, opened 2021-07-23
fix(n-tree-select): prop checkable doesn't work

kungege:
fix [#618](https://github.com/TuSimple/naive-ui/issues/618)


amadeus711:
Sorry, this is NTreeSelect's demo bug, you can refer to https://codesandbox.io/s/14318

PR #643, opened 2021-07-25
fix(MessageProvider): remove unnecessary import

kungege:


kungege:
@07akioni [here](https://github.com/TuSimple/naive-ui/runs/3153705546#step:5:14)

07akioni:
已经 fix 了

PR #695, opened 2021-07-28
feat(NTimePicker): NTimePicker optimization the now button logic

### LINKED ISSUE #401: allow to remove the `now` button from `n-time-picker`
to use the timepicker with disabled values. In that case the `now` button is useless and can be confusing for users
### Expected API (期望的 API)
maybe something similar to what has been done with `n-date-picker` `actions` prop?

amadeus711:
fix #401

### DIFF
--- a/src/time-picker/src/utils.ts
+++ b/src/time-picker/src/utils.ts
@@ -204,7 +204,7 @@ export function findSimilarTime (
   if (lowerBound === undefined) {
     if (!upperBound) {
       throwError(
-        'n-time-picker',
+        'time-picker',
         "Please set 'hours' or 'minutes' or 'seconds' props"
       )
     }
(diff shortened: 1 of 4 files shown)


src/time-picker/src/utils.ts
```diff
@@ -184,3 +184,31 @@ export function isTimeInStep (
...
+    if (v === value) return v
+    else if (v > value) {
+      upperBound = v
+      break
+    }
+    lowerBound = v
+  }
+  if (lowerBound === undefined) {
+    if (!upperBound) {
+      throw new Error("Please set 'hours' or 'minutes' or 'seconds' props")
```
07akioni, line 206: 这个地方用 throwError

src/time-picker/src/utils.ts
```diff
@@ -203,7 +203,10 @@ export function findSimilarTime (
   }
   if (lowerBound === undefined) {
     if (!upperBound) {
-      throw new Error("Please set 'hours' or 'minutes' or 'seconds' props")
+      throwError(
+        'n-time-picker',
```
07akioni, line 207:
```suggestion
        'time-picker',
```

PR #847, opened 2021-08-11
[WIP] feat(message): Add `cssClassWrapper` prop to `Message`

rhengles:
### Work in Progress !

Before writing tests and docs, I want to reach an agreement with the contributors if this feature and specific implementation is desirable.

This allows one to customize the style of individual messages, passing one or more custom css classes to be applied to individual messages, instead of styling them all.

Ideally I would like to be able to use completely custom elements, without requiring wrapper elements like `<div class="n-message-wrapper">` and `<div class="n-message">`.

### DIFF
--- a/src/message/src/Message.tsx
+++ b/src/message/src/Message.tsx
@@ -117,11 +117,12 @@ export default defineComponent({
       content,
       mergedClsPrefix,
       cssVars,
-      handleClose
+      handleClose,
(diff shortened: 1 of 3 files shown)


rhengles:
I've had a better idea, I will submit another PR with a simpler and more powerful alternative, you can consider to close this if you wish

PR #855, opened 2021-08-12
feat(message): Add `raw` prop to `Message`

rhengles:
### Work in Progress

If this feature and implementation is approved, I can work on writing tests and documentation :)

This adds a `raw` property on `NMessage`, allowing one to render completely custom elements inside of a message, without the frame and icon. But this allows one to render custom content along with other messages, without interfering/hiding them.

### DIFF
--- a/src/message/src/Message.tsx
+++ b/src/message/src/Message.tsx
@@ -119,7 +119,9 @@ export default defineComponent({
       cssVars,
       handleClose
     } = this
-    return (
+    return this.raw ? (
+      render(content)
+    ) : (
       <div
         class={`${mergedClsPrefix}-message-wrapper`}
         style={cssVars as CSSProperties}
--- a/src/message/src/MessageEnvironment.tsx
+++ b/src/message/src/MessageEnvironment.tsx
@@ -81,6 +81,7 @@ export default defineComponent({
                 content={this.content}
                 type={this.type}
                 icon={this.icon}
+                raw={this.raw}
                 closable={this.closable}
                 onClose={this.handleClose}
               />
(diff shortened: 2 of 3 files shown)


rhengles:
This replaces PR #847

07akioni:
closed in 93c4a4cd3695fa0cd5c510da014bfe389467aa1f

rhengles:
Awesome! Thanks 💯🎉 You rock! 👊

rhengles:
@07akioni 

Just two questions I have, they're somewhat related:

1. Sometimes I need to use the NAlert, but sometimes I need to use another component to render the Message. I know this function can handle it, I'm just in doubt about where can I store this information so the render function can get it. Maybe if I could define this `renderMessage` function _per message_ instead of _globally,_ that would be perfect.

2. Is there a way for me to use this `renderFunction`, but conditionally be able to use the default NMessage panel ?

Thanks again.

07akioni:
> @07akioni
> 
> Just two questions I have, they're somewhat related:
> 
> 1. Sometimes I need to use the NAlert, but sometimes I need to use another component to render the Message. I know this function can handle it, I'm just in doubt about where can I store this information so the render function can get it. Maybe if I could define this `renderMessage` function _per message_ instead of _globally,_ that would be perfect.
> 2. Is there a way for me to use this `renderFunction`, but conditionally be able to use the default NMessage panel ?
> 
> Thanks again.

1. In the current API you may need to break some typescript rules to pass invalid type such as `alert:xxx` and consume.
2. Impossible for now, I may have a try to make it work.

PR #885, opened 2021-08-15
docs(NAvatar, NButton): Avatar and button page improvements

xon52:


xon52:
Planning on going through all the English docs like this, but just checking it's what you want before I do that - so please let me know if there's anything you don't like!

amadeus711:
> Planning on going through all the English docs like this, but just checking it's what you want before I do that - so please let me know if there's anything you don't like!

Welcome your contribution, but this pr is about docs, please checkout from `docs` branch and set `docs` branch as the target branch.

amadeus711 (approved):

PR #893, opened 2021-08-16
Docs(NCarousel) Remove autoplay from non-autoplay examples

xon52:
I found the autoplaying confusing as to what was different between these.
I think the autoplay should be left for only the autoplay example so you can explore the specific differences of the other examples.

### DIFF
--- a/src/carousel/demos/enUS/dot-placement.demo.md
+++ b/src/carousel/demos/enUS/dot-placement.demo.md
@@ -3,7 +3,7 @@
 Set `dot-placement` to change the indication point position.
 
 ```html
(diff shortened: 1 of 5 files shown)


amadeus711:
I think the example should indicate that users can try as many combinations as possible, rather than controlling variables to find different ones. Sorry, my English is not very good. I'm not sure if you understand me.

xon52:
> I think the example should indicate that users can try as many combinations as possible, rather than controlling variables to find different ones. Sorry, my English is not very good. I'm not sure if you understand me.

Your English is good 👍 
I understand your concerns, but I see akioni already merged this. For this example though I found the autoplay to be very distracting. Maybe we could have other combinations that do not include autoplay?

PR #925, opened 2021-08-20
test(transfer): add test

amadeus711:

### DIFF
--- a/src/transfer/tests/Transfer.spec.ts
+++ b/src/transfer/tests/Transfer.spec.ts
@@ -1,8 +1,39 @@
 import { mount } from '@vue/test-utils'
+import { sleep } from 'seemly'
 import { NTransfer } from '../index'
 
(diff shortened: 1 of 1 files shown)


src/transfer/tests/Transfer.spec.ts
```diff
@@ -1,8 +1,39 @@
...
+        value: 'test1'
+      }
+    ]
+    const onFilter = jest.fn()
+    const wrapper = mount(NTransfer, {
+      props: { filterable: true, filter: onFilter, options: options }
+    })
+    await wrapper.find('input').setValue('1')
+    await sleep(300)
+    expect(onFilter).toHaveBeenCalled()
```
07akioni, line 37: why sleep(300)?

Use a proper way.

07akioni (commented):
Change `sleep(300)`

PR #961, opened 2021-08-25
fix(dialog):'width:446px' to 'min-width: 446px'

daipeng7:


07akioni:
cascader 修复已经合并了，这个我先关了

PR #1020, opened 2021-08-31
docs(anchor): Upgrade the use case of anchor component to Composi…

songjianet:
…tion API


songjianet:
commit message写错了，不好意思

amadeus711 (approved):

PR #42, opened 2021-06-09
feat(dropdown): add disabled prop

ahwgs:

PR #1040, opened 2021-09-01
feat(tree&tree-select): add children-field prop

kungege:
ref [#42](https://github.com/07akioni/treemate/pull/42)


07akioni:
@kungege `treemate@0.3.3`

PR #1057, opened 2021-09-02
chore: update deprecated deps

equt:
In this PR,

1. deprecated packages are replaced by the successors, e.g., `babel-eslint`;
2. missing peer deps are added;
3. orphan deps are removed, e.g. `eslint-plugin-standard`.

This PR should break nothing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,8 @@
     ]
   },
   "devDependencies": {
+    "@babel/core": "^7.15.0",
+    "@babel/eslint-parser": "^7.15.0",
     "@babel/generator": "^7.12.11",
     "@babel/parser": "^7.12.11",
     "@babel/preset-env": "^7.12.10",
@@ -82,7 +84,6 @@
     "@vue/server-renderer": "^3.0.11",
     "@vue/test-utils": "^2.0.0-rc.9",
     "autoprefixer": "^10.2.6",
-    "babel-eslint": "^10.1.0",
     "babel-jest": "^27.0.2",
     "codesandbox": "^2.2.3",
     "cross-env": "^7.0.3",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -94,7 +95,6 @@
     "eslint-plugin-markdown": "^2.0.0",
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-promise": "^5.1.0",
-    "eslint-plugin-standard": "^5.0.0",
```
07akioni, line 97: If this is removed, what can make the code follows standard style?

package.json
```diff
@@ -103,13 +103,16 @@
...
     "prettier": "^2.2.1",
     "rimraf": "^3.0.2",
+    "rollup": "^2.56.3",
     "superagent": "^6.1.0",
     "typescript": "~4.3.0",
     "vite": "^2.1.3",
     "vue": "^3.2.1",
-    "vue-router": "^4.0.5"
+    "vue-router": "^4.0.5",
+    "webpack": "^5.51.1"
```
07akioni, line 115: webpack is not mean to be a dev dep. it is only needed when user want to use playground (i want them to install webpack manually, since it is not small)

package.json
equt: According to the deprecation message [here](https://www.npmjs.com/package/eslint-plugin-standard),

> eslint-config-standard 16.0.0 no longer require the eslint-plugin-standard package

package.json
equt: This is reasonable, let's see if we can make it optional.

equt: Well, this is not possible, but I'm ok with one more dev dep, comparing to the annoying warning message from the package manager.

07akioni: playground is not for daily developing but for experiments. adding webpack to dev dep wont make examples inside playground work out of box. so just remove it, for people
who want to do experiments, they can install webpack or
 express manually.

07akioni: Webpack is not required for component developing.

equt: That's fair.

equt:
I accidentally deleted my fork earlier, and there seems to have no way to relink this PR to my fork. Should I just close this one in favor of a new one?

07akioni:
> y fork earlier, and there seems to have no way to relink this PR to my fork. Should I just close this one in favor of a new one?

OK.

Note that npm7 will install peerDeps automatically, I think add missing peers is not that necessary.

PR #1058, opened 2021-09-02
chore: let eslint respect prettier config

equt:
This PR adds `eslint-config-prettier` to make sure ESLint respects Prettier's config.

And this would fix the problem like:

Code Diff


equt:
BTW, lint still has to be ran after format, since ESLint will change the code layout in the setup.

kungege:
LGTM

PR #1148, opened 2021-09-09
Update basic.demo.md

waiylgeek:
Prevent from circling the steps and define number of steps


07akioni:
This is  by-design to test the case.

PR #1206, opened 2021-09-20
fix(n-tree): check node does not work when setting cascade is false

LYErin:


src/tree/demos/zhCN/cascade.demo.md
```diff
@@ -1,6 +1,6 @@
 # 级联选择
 
-设定 `cascade` 进行级联选择。
+设定 `cascade` 为 false 进行非级联选择。
```
07akioni, line 3:
```suggestion
设定 `cascade` 进行级联选择。
```

src/tree/demos/enUS/cascade.demo.md
```diff
@@ -1,6 +1,6 @@
 # Cascade Check
 
-Set `cascade` to use cascade check.
+Set `cascade` is false to use noncascade check.
```
07akioni, line 3:
```suggestion
Set `cascade` to use cascade check.
```

Issue #1211, opened 2021-09-21
Drawer header or footer slot missing borderColor transition 


zmtlwzy:
### TuSimple/naive-ui version (版本)
2.19.1

### Vue version (Vue 版本)
3.2.12

### Browser and its version (浏览器及其版本)
Edge(latest)

### System and its version (系统及其版本)
codesandbox

### Node version (Node 版本)
latest

### Reappearance link (重现链接)
https://codesandbox.io/s/silly-hill-qz20b?file=/src/Demo.vue

### Reappearance steps (重现步骤)
Drawer header or footer slot missing borderColor transition ,
When the drawer component has a component for switching darkmode.

### Expected results (期望的结果)
show transition when switch darkmode

### Actual results (实际的结果)
no transition when switch darkmode

PR #1213, opened 2021-09-21
Fix: Add border transition on the drawer

Matth10:
### Description
This branch is aiming to fix this issue #1211. It simply adds some border transition on the header and footer of the drawer.

### With the transition
Kapture 2021-09-21 at 19 54 22


07akioni:
please add changelog

Issue #41, opened 2021-06-09
Feature request: better ARIA support and less click listeners


gdh1995:
## What problem does the feature solve?

I find there're two issues on the doc website:
* some clickable targets have no `[role="button"]`, like dropdown menuitems, radio buttons and breadcrumbs
* some `disabled` components still has click listeners but no `[aria-disabled]` is added
  * like tags, cascader menuitems, checkboxes
* usually a `<input>` is wrapped by `<div>`s, and the `div.n-input` has a click listener but without `[aria-hidden=true]`

## What does the proposed API look like?

I have an extension "[Vimium C](https://github.com/gdh1995/vimium-c)" to help users use keyboard to click page elements, and when it find links, it will pick those with `[role=button], [role=link]` and filter out those `[aria-hidden], [aria-disabled], [aria-hidden=true], [aria-disabled=true]`. It also hooks the global `addEventListener` function to learn what "plain" elements are clickable.

When I tested Vimium C's `LinkHints.activate` (triggered by `f`) on pages like https://www.naiveui.com/zh-CN/os-theme/components/input , I found many clickable were not recognized, while many others were mistakenly hinted.

So I want this framework to add more clues to let such keyboard helper extensions work better.

07akioni:
I'm going to add better aria support. Actually I'm new to a11y. The support of a11y will be gradually added.

You advice is very useful. I'll keep the issue open. If anyone could help I'd very appreciate.

PR welcomed!

timeTravelCYN:
> I'm going to add better aria support. Actually I'm new to a11y. The support of a11y will be gradually added.
> 
> You advice is very useful. I'll keep the issue open. If anyone could help I'd very appreciate.
> 
> PR welcomed!

@07akioni 🙋‍♂️

delowardev:
Any update on this ticket? I'm also thinking of implementing a11y and better keyboard support.

07akioni:
> t would be helpful if you mention which component you're working on so that we can start with a different component.

Currently you can pick any components you want to implement. Just mark it here. If it's duplicate I've let you known.

07akioni:
> Any update on this ticket? I'm also thinking of implementing a11y and better keyboard support. It would be helpful if you mention which component you're working on so that we can start with a different component.

Currently you can pick any component you want to implement. Just mark it here. If it's duplicate I'll let you know.

Matth10:
[...]

PR #1226, opened 2021-09-23
Accessibility/improve aria for checkbox

Matth10:
### Description
Add better aria-support for `n-checkbox` and `n-checkbox-group`. Followed this guide w3.org/TR/wai-aria-practices-1.1/examples/checkbox/checkbox-1/checkbox-1.html.

#### NCheckboxGroup
- Added a `role="group"`

#### NCheckbox
- Added a `role="checkbox"`
- Added `aria-labelledby="some-if"`. By default, we will give a custom id (Ex: `n-checbox-label-CBMZ`) to the label wrapper so we can reference this label in the aria attribute.
- Added `aria-checked` with the value of `true` or `false` and `mixed` if the checkbox is in an intermediate state.

### Result
#41 . Now it should be better for the browser to understand the checkbox input.
Screen Shot 2021-09-23 at 10 37 13 AM

### Consideration
One thing we can do is to refactor the component to use `<input/>` and `<label>` to have a better semantic and some accessibility features out of the box.
Like something like this for example
```html
<div>
  <div class="hidden-accessible">
     <input type="checkbox" value="Chicago" id="city1" name="city">
  </div>
  <div class="checkbox-box" role="checkbox" aria-checked="false">
    <span class="checkbox-icon"></span>
  </div>
  <label for="city1">Chicago</label>
</div>
```

### DIFF
--- a/src/checkbox/src/Checkbox.tsx
+++ b/src/checkbox/src/Checkbox.tsx
@@ -285,7 +285,7 @@ export default defineComponent({
         ]}
         tabindex={mergedDisabled || !focusable ? undefined : 0}
         role="checkbox"
-        aria-checked={renderedChecked}
+        aria-checked={indeterminate ? 'mixed' : renderedChecked}
         style={cssVars as CSSProperties}
         onKeyup={handleKeyUp}
         onKeydown={handleKeyDown}
--- a/src/checkbox/src/Checkbox.tsx
+++ b/src/checkbox/src/Checkbox.tsx
@@ -285,6 +285,7 @@ export default defineComponent({
         ]}
         tabindex={mergedDisabled || !focusable ? undefined : 0}
         role="checkbox"
+        aria-checked={renderedChecked}
         style={cssVars as CSSProperties}
         onKeyup={handleKeyUp}
         onKeydown={handleKeyDown}
--- a/src/checkbox/src/Checkbox.tsx
+++ b/src/checkbox/src/Checkbox.tsx
@@ -10,6 +10,7 @@ import {
   CSSProperties
 } from 'vue'
 import { useMergedState, useMemo } from 'vooks'
+import { createId } from 'seemly'
 import { useConfig, useFormItem, useTheme } from '../../_mixins'
 import type { ThemeProps } from '../../_mixins'
 import { NIconSwitchTransition } from '../../_internal'
@@ -153,6 +154,7 @@ export default defineComponent({
       props,
       mergedClsPrefixRef
     )
+
     function toggle (e: MouseEvent | KeyboardEvent): void {
       if (NCheckboxGroup && props.value !== undefined) {
         NCheckboxGroup.toggleCheckbox(!renderedCheckedRef.value, props.value)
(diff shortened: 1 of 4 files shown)


src/_utils/generators/id.ts
```diff
@@ -0,0 +1,8 @@
+export function generateId (prefix: string, length: number = 4): string {
+  let uuid = ''
+  const dictionary = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz'
+  for (let i = 0; i < length; i++) {
+    uuid += dictionary.charAt(Math.floor(Math.random() * dictionary.length))
+  }
+  return `${prefix}-${uuid}`
+}
```
07akioni, line 8:
```
import { createId } from 'seemly'
```

src/checkbox/src/Checkbox.tsx
```diff
@@ -153,6 +154,9 @@ export default defineComponent({
       props,
       mergedClsPrefixRef
     )
+    const labelId = computed(() => generateId('n-checbox-label'))
```
07akioni, line 157: use `createId` from 'seemly'.

src/checkbox/src/Checkbox.tsx
```diff
@@ -153,6 +154,9 @@ export default defineComponent({
...
+    const ariaLabelledBy = computed(() => labelId.value)
```
07akioni, line 158: No need for reactivity, just create a const and expose it to template.

The value won't change in the entire component life cycle.

src/_utils/generators/id.ts
07akioni: No need for the file

07akioni:
It's possible to add a invisible `input` element. Will it be hard? At least the focus related code need to be refactored (like radio).

src/_utils/generators/id.ts
Matth10: perfect I didn't know you had a lib for that

src/checkbox/src/Checkbox.tsx
```diff
@@ -197,6 +199,7 @@ export default defineComponent({
       mergedDisabled: mergedDisabledRef,
       renderedChecked: renderedCheckedRef,
       mergedTheme: themeRef,
+      labelId: `n-checbox-label-${createId(4)}`,
```
07akioni, line 202: The prefix may not be needed?

Matth10: Oh and I made a typo. I don't know we don't need it but might be more user friendly if we want to search for it in the devtool console

Matth10:
> It's possible to add an invisible `input` element. Will it be hard? At least the focus related code need to be refactored (like radio).

I don't think it will be hard. We can use some CSS to hide the input visually. Instead of using a `tabindex=0` on the checkbox wrapper we can have a `focus` event handler on the input and change the style of the wrapper when the input is focused or not.

I don't know if you want me to try on this branch or create another PR?

07akioni:
> > It's possible to add an invisible `input` element. Will it be hard? At least the focus related code need to be refactored (like radio).
> 
> I don't think it will be hard. We can use some CSS to hide the input visually. Instead of using a `tabindex=0` on the checkbox wrapper we can have a `focus` event handler on the input and change the style of the wrapper when the input is focused or not.
> 
> I don't know if you want me to try on this branch or create another PR?

Maybe in another PR.

PR #1248, opened 2021-09-27
Accessibility/improve aria support breadcrumb

Matth10:
### Description
#41 This PR is aiming to improve the accessibility of the `NBreadcrumb` component. Followed: https://www.w3.org/TR/wai-aria-practices-1.1/examples/breadcrumb/index.html

### Problems identified
- Missing `nav` tag with a `aria-label="Breadcrumb"`
- Missing a list tag to make it easier for the browser to understand the list of breadcrumb items
- Telling the currently active link

### Implmentation and changes
Most of the time each breadcrumb items is a link. Using a slot to let the user pass anything is a good idea especially if the they want to use a `router-link` or a `dropdown`. However I think we should let the user pass an `href` so we can construct a link tag from it very easily. 

Before:

```jsx
<NBreadCrumb>
   <NBreadCrumbItem>
        <a href="#">This is a basic link</a>
   </NBreadCrumbItem>
</NBreadCrumb>
```

Instead having that:

```jsx
<NBreadCrumb>
   <NBreadCrumbItem href="#">
        This is a basic link
   </NBreadCrumbItem>
</NBreadCrumb>
```

Thanks to this we can apply automatically the `aria-current` if the item is the current location. However if the user decides to use the slot with a custom link or dropdown I think it's their responsibility to place the `aria-current` in the right place. But it is nice to have an easy way to create a link.

What are your thoughts on this?

### Documentation
I updated the US doc. If we go with this API (no breaking changes) I might need some help for the Chinese version 😄 . Also I changed the dropdown section to be `custom` so we can show how to pass anything not just a dropdown. (I think `router-link` is what we will pass most of the time`)


src/breadcrumb/src/BreadcrumbItem.tsx
```diff
@@ -13,7 +22,7 @@ ExtractPropTypes<typeof breadcrumbItemProps>
 export default defineComponent({
   name: 'BreadcrumbItem',
   props: breadcrumbItemProps,
-  setup (props, { slots }) {
+  setup ({ separator, href, isCurrent = false }, { slots }) {
```
07akioni, line 25: You can't do destructing in setup since the data would lost reactivity.

src/breadcrumb/src/BreadcrumbItem.tsx
```diff
@@ -25,20 +34,31 @@ export default defineComponent({
...
-            {slots}
-          </span>
-          <span class={`${mergedClsPrefix}-breadcrumb-item__separator`}>
-            {slots.separator
-              ? slots.separator()
-              : props.separator ?? separatorRef.value}
-          </span>
-        </span>
-      )
+
```
07akioni, line 42: Using JSX mixed with render function is a more elegant way.

```
<li class={`${mergedClsPrefix}-breadcrumb-item`}>
  {h.....}
  <span>...</span>
</li>
```

src/breadcrumb/src/BreadcrumbItem.tsx
```diff
@@ -25,20 +34,31 @@ export default defineComponent({
       return () => null
     }
     const { separatorRef, mergedClsPrefixRef } = NBreadcrumb
+    const htmlTag = computed(() => (href ? 'a' : 'span'))
+    const ariaCurrent = computed(() => (href && isCurrent ? 'location' : null))
```
07akioni, line 38: I think we should check window location instead and don't need user to pass this prop.
In most cases user would just do the same thing.

src/breadcrumb/src/BreadcrumbItem.tsx
07akioni: For example if separator changes the item won't rerender.

Matth10: Ok! I'm new to vue so I might make some dumb mistake like this :)

src/breadcrumb/src/BreadcrumbItem.tsx
Matth10: Yes, I think 90% of the case the devs are using a lib for the routing (vue-router) and use their own link component that handles this automatically (`aria-current`). But yes we can remove this prop to handle that automatically if the user passes an `href`

src/breadcrumb/src/BreadcrumbItem.tsx
07akioni: No problem, you'll be more and more experienced.

src/breadcrumb/src/BreadcrumbItem.tsx
07akioni: Yep, I think it's meaningless to set `isCurrent = true` if anchor's href doesn't match the current location and vice versa.

src/breadcrumb/tests/Breadcrumb.spec.ts
```diff
@@ -1,12 +1,23 @@
 import { h } from 'vue'
 import { mount } from '@vue/test-utils'
 import { NBreadcrumb, NBreadcrumbItem } from '../index'
+import Breadcrumb from '../src/Breadcrumb'
```
07akioni, line 4: Should this be `NBreadcrumb` on line 3?

src/_utils/composable/use-browser-location.ts
```diff
@@ -0,0 +1,67 @@
...
+  hostname?: string
+  href?: string
+  origin?: string
+  pathname?: string
+  port?: string
+  protocol?: string
+  search?: string
+}
+
+export const useBrowserLocation = (
```
Matth10, line 18: Use this composable to make the window location reactive

src/breadcrumb/src/BreadcrumbItem.tsx
Matth10: One possible problem of this approach (`window.location`) is we have now the responsibility to handle location changes and set the right aria attrs depending on that. This kind of increases the complexity of this component. 
I made a possible implementation of this. But I'm wondering if it's worth handling that compare to having a prop (`isCurrent`). Especially because these props `href` and `isCurrent` won't be used that much as the user will use `router-link` or `nuxt-link` most of the time.

src/breadcrumb/src/BreadcrumbItem.tsx
```diff
@@ -25,19 +34,36 @@ export default defineComponent({
       return () => null
     }
     const { separatorRef, mergedClsPrefixRef } = NBreadcrumb
+    const browserLocation = useBrowserLocation()
+
+    const htmlTag = computed(() => (props.href ? 'a' : 'span'))
+    const ariaCurrent = computed(() =>
```
07akioni, line 40:
```suggestion
    const htmlTagRef = computed(() => (props.href ? 'a' : 'span'))
    const ariaCurrentRef = computed(() =>
```

src/breadcrumb/src/BreadcrumbItem.tsx
```diff
@@ -1,9 +1,18 @@
...
+  computed
+} from 'vue'
 import { warn } from '../../_utils'
+import { useBrowserLocation } from '../../_utils/composable/use-browser-location'
 import { breadcrumbInjectionKey } from './Breadcrumb'
 
 const breadcrumbItemProps = {
-  separator: String
+  separator: String,
+  href: String as PropType<string | undefined>
```
07akioni, line 15:
```suggestion
  href: String
```

`PropType` is not needed, the inferred type would be the same.

src/breadcrumb/src/BreadcrumbItem.tsx
```diff
@@ -25,19 +34,36 @@ export default defineComponent({
       return () => null
     }
     const { separatorRef, mergedClsPrefixRef } = NBreadcrumb
+    const browserLocation = useBrowserLocation()
```
07akioni, line 37:
```suggestion
    const browserLocationRef = useBrowserLocation()
```

src/_utils/composable/use-browser-location.ts
```diff
@@ -0,0 +1,67 @@
+import { onMounted, onUnmounted, Ref, ref } from 'vue'
+
+const hasWindow = typeof window !== 'undefined'
+const defaultWindow = hasWindow ? window : null
```
07akioni, line 4:
```suggestion
const defaultWindow = typeof window !== 'undefined' ? window : null
```

Matth10: I think we should keep this function `hasWindow` and external it into its own utils module. So we can use it when we want to interact with the `window` object.

PR #1277, opened 2021-10-04
feat(n-tree): n-tree's on-update prop add option info.

### LINKED ISSUE #1274: tree's 'update-selected-keys' event should pass the node item as parameter
now in 'update-selected-keys' event, user can only get key, and have to iterate whole nodes to get other property, or store it at beginning
### Expected API (期望的 API)
onUpdateSelectedKeys?: MaybeArray<(value: Key[],nodes:Any[]) => void> | undefined;

amadeus711:
close #1274

### DIFF
--- a/src/tree/src/Tree.tsx
+++ b/src/tree/src/Tree.tsx
@@ -79,10 +79,7 @@ export function createTreeMateOptions<T> (
   }
 }
 
-type OnUpdateCallback = (
-  value: Key[],
-  option: Array<TreeOption | null> | null
-) => void
+type OnUpdateCallback = (value: Key[], option: Array<TreeOption | null>) => void
 
 export const treeSharedProps = {
   filter: Function as PropType<(pattern: string, node: TreeOption) => boolean>,
@@ -521,7 +518,7 @@ export default defineComponent({
 
     function doUpdateExpandedKeys (
       value: Key[],
-      option: Array<TreeOption | null> | null
+      option: Array<TreeOption | null>
     ): void {
       const {
         'onUpdate:expandedKeys': _onUpdateExpandedKeys,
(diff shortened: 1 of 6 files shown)


07akioni:
@amadeus711 rebase 一下吧

src/tree/src/Tree.tsx
```diff
@@ -101,16 +103,16 @@ export const treeSharedProps = {
   },
   indeterminateKeys: Array as PropType<Key[]>,
   onUpdateIndeterminateKeys: [Function, Array] as PropType<
-  MaybeArray<(value: Key[]) => void>
+  MaybeArray<OnUpdateCallback>
```
07akioni, line 106:
```suggestion
  MaybeArray<OnUpdateKeys>
```

src/tree/src/Tree.tsx
```diff
@@ -506,40 +511,61 @@ export default defineComponent({
       }
     }
 
-    function doUpdateExpandedKeys (value: Key[]): void {
+    function getOptionsByKeys (keys: Key[]): Array<TreeOption | null> {
+      const getNode = dataTreeMateRef.value!.getNode
```
07akioni, line 515:
```suggestion
      const { getNode } = dataTreeMateRef.value!
```

src/tree/src/Tree.tsx
```diff
@@ -692,7 +725,8 @@ export default defineComponent({
           droppingMouseNode.key === node.key &&
           !mergedExpandedKeysRef.value.includes(node.key)
         ) {
-          doUpdateExpandedKeys(mergedExpandedKeysRef.value.concat(node.key))
+          const newKeys = mergedExpandedKeysRef.value.concat(node.key)
```
07akioni, line 728:
```suggestion
          const nextKeys = mergedExpandedKeysRef.value.concat(node.key)
```

PR #1522, opened 2021-11-04
fix(ci): set NODE_OPTIONS=--unhandled-rejections=warn

Sepush:
set `NODE_OPTIONS=--unhandled-rejections=warn` to add node 16 test


07akioni:
原来是这个原因，😅

PR #1540, opened 2021-11-08
feat(menu): add a color distinction between selected and unselected arrow

### LINKED ISSUE #1535: Adds a color distinction between selected and unselected ICONS for menu components
Adds a color distinction between selected and unselected ICONS for menu components
image

### Expected API (期望的 API)
Currently, both checked and unchecked ICONS look checked

LYErin:
close #1535

### DIFF
--- a/CHANGELOG.en-US.md
+++ b/CHANGELOG.en-US.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## Pending
+
+### Fixes
+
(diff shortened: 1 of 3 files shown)


Talljack:
The icon color is not activated when active. I think this is needed.

LYErin:
> The icon color is not activated when active. I think this is needed. image

The arrow will set color just on selected and hovered.
image

amadeus711 (approved):

Talljack:
> > The icon color is not activated when active. I think this is needed. image
> 
> The arrow will set color just on selected and hovered. image

OK

PR #1591, opened 2021-11-15
Feature/data table striped

JerryChan31:
implementation of https://github.com/TuSimple/naive-ui/issues/1552.

Something may need to be discussed:
1. I'm not sure if it's proper to use `var(--td-color-hover)` as stripe color.
2. I'm using just css to implement this feature, but it seems not possible to test it with Jest. Is there a better way?

It's my first PR to open source community so if there's any problem just point it out and I'll fix it.


src/data-table/src/styles/index.cssr.ts
```diff
@@ -303,6 +303,9 @@ export default c([
     cM('loading', [
       cB('data-table-wrapper', 'opacity: var(--opacity-loading);')
     ]),
+    cM('striped', [
+      cB('data-table-tr:nth-child(even)', [cB('data-table-td', 'background-color: var(--td-color-hover);')])
+    ]),
```
07akioni, line 308: In fact this is not the correct behavior since when table row is expandable, the even odd status of the row will be changed.

However it won't be hard since we can just follow the index to distinguish the row status.

07akioni:
I'm wrong, actually the refactor is not easy. hahaha.

See c522347515b19bf502cc8de6418f30537cf4edd7 to know what is changed to make it work in all possible cases.

PR #1604, opened 2021-11-17
fix(button): not not click when loading

### LINKED ISSUE #1601: You should not be able to click while the button is loaded
You should not be able to click while the button is loaded
### Expected API (期望的 API)
You should not be able to click while the button is loaded

image
image

LYErin:
close #1601


07akioni:
感觉 loading 也不影响点击

jahnli:
@07akioni   when I click the button in loading, I will always request events

Issue #636, opened 2021-07-24
Nuxtjs Support


pipinet:
### This function solves the problem (这个功能解决的问题)
Nuxtjs Support
### Expected API (期望的 API)
Nuxtjs Support

07akioni:
我不熟这个，欢迎有兴趣的人提 PR

I'm not familar with it, PR welcomed

liaozheng-yi:
据我所知，Nuxtjs目前是基于vue2的，而vue3驱动的Nuxt3还在开发中，所以，同样基于vue3的naive-ui很难支持当下的Nuxtjs。

As I know, Nuxtjs is based on vue2, and the Nuxt3 powered with vue3 still in heavy development. So,  navie-ui that is also based on vue3 is hard to surpport Nuxtjs now.

07akioni:
> 据我所知，Nuxtjs目前是基于vue2的，而vue3驱动的Nuxt3还在开发中，所以，同样基于vue3的naive-ui很难支持当下的Nuxtjs。
> 
> As I know, Nuxtjs is based on vue2, and the Nuxt3 powered with vue3 still in heavy development. So, navie-ui that is also based on vue3 is hard to surpport Nuxtjs now.

我不熟 NUXT，不过估计主要的难点在于 css-in-js 导致的 SSR 问题，其他没啥道理不 work。

pipinet:
https://nuxtjs.org/v3

migueldaipre:
> If someone has tried it, please tell us what errors you met.
> 
> Actually I've never used nuxt and don't know what's required to make it work.

I did the default installation steps for nuxt ([https://v3.nuxtjs.org/getting-started/installation](url)) and for naive-ui ([https://www.naiveui.com/en-US/os- theme/docs/installation](url)) and I'm getting the error "Server Side Rendering Error: Error: Unexpected token 'export'", follow StackBlitz link to help debug [https://stackblitz.com/edit/github-ovqtjv?devtoolsheight= 33&file=app.vue](url). thanks

07akioni:
That's quite wired since we do provide a cjs build version in `node_modules/naive-ui/lib/`

07akioni:
> I'm interested in getting involved in the job, but I'm new if I can get advice on where to start.

Just try to make a button rendered in nuxt.

pipinet:
[...]

PR #1611, opened 2021-11-17
Update styles to bypass Vitejs bug on string interpolation

jbmolle:
Hi,
There is a bug on Vitejs when using a javascript object in string interpolation.
Therefore when building Nuxt app with Vitejs, we got an undefined commonVariables.
This PR solves this problem and should solve #636 
In defineNuxtConfig function we don't need to add vite: false anymore.


notbucai (commented):
It looks like it's just deconstructing the object。

PR #1620, opened 2021-11-19
Fix(dropdown) transition shifts in cascade dropdown with anchor option

doom-9:


doom-9:
close https://github.com/TuSimple/naive-ui/issues/1600

PR #1635, opened 2021-11-21
feat(slider): support `mark`

nooooooom:
`step` prop support `mark` option.


src/slider/demos/enUS/restrict-selectable-values.demo.md
```diff
@@ -0,0 +1,27 @@
+# Restrict Selectable Values
```
07akioni, line 1:
```suggestion
# Restrict selectable values
```

PR #1654, opened 2021-11-22
fix(slider): tooltip tracking timing

nooooooom:

### DIFF
--- a/src/slider/src/Slider.tsx
+++ b/src/slider/src/Slider.tsx
@@ -211,7 +211,7 @@ export default defineComponent({
       return mergedMarks
     })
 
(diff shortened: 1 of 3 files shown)


src/slider/src/Slider.tsx
```diff
@@ -714,6 +720,7 @@ export default defineComponent({
                           ref={this.setFollowerRefs(index)}
                           show={showTooltip}
                           to={this.adjustedTo}
+                          enabled={this.followerEnabledIndex.has(index)}
```
07akioni, line 723: followerEnabledIndexSet

PR #1692, opened 2021-11-27
fix(log): `trim` prop not being independent when used

amadeus711:

### DIFF
--- a/src/log/src/LogLine.tsx
+++ b/src/log/src/LogLine.tsx
@@ -1,4 +1,13 @@
-import { h, defineComponent, inject, ref, onMounted, watch, toRef } from 'vue'
+import {
+  h,
+  defineComponent,
+  inject,
+  ref,
+  onMounted,
+  watch,
+  toRef,
+  computed
+} from 'vue'
 import { logInjectionKey } from './Log'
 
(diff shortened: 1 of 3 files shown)


src/log/src/LogLine.tsx
```diff
@@ -9,27 +18,25 @@ export default defineComponent({
...
-    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    const { trimRef, highlightRef, languageRef, mergedHljsRef } = inject(
-      logInjectionKey
-    )!
+    const { trimRef, highlightRef, languageRef, mergedHljsRef } =
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      inject(logInjectionKey)!
     const selfRef = ref<HTMLElement | null>(null)
+    const mergedLineTrimRef = computed(() => {
+      return trimRef.value ? (props.line || '').trim() : props.line
```
07akioni, line 26:
```suggestion
      return trimRef.value ? (props.line || '').trim() : props.line
```

`(props.line || '')` 没必要

07akioni: 虽然之前也是我写的= =

PR #1720, opened 2021-12-02
fix(dialog): `content` style problem when use English

amadeus711:


src/dialog/src/styles/index.cssr.ts
```diff
@@ -77,6 +77,7 @@ export default c([
       font-size: var(--font-size);
       margin: var(--content-margin);
       position: relative;
+      word-break:break-all;
```
07akioni, line 80:
```suggestion
      word-break: break-word;
```

PR #1778, opened 2021-12-07
docs: Add missing types to 'preset' of Modal props

OrbisK:


07akioni:
Thanks for fixing that, in fact confirm dialog should be deprecated.

PR #1804, opened 2021-12-09
docs(page-header): use ellipsis icon in template instead of text '...'

jdwit:


07akioni:
Sorry for that but text is expected since the icon is too strong for the button compare with text stroke.

jdwit:
> Sorry for that but text is expected since the icon is too strong for the button compare with text stroke.

I agree text looks better but I figured the icon was expected since the `EllipsisHorizontal` is imported but not used in the template. Maybe remove the import statement?

Issue #1842, opened 2021-12-14
Tracking issue for let CSS var prefixed with n


Sepush:
### This function solves the problem (这个功能解决的问题)
solve a kind of problem like https://github.com/TuSimple/naive-ui/issues/1808
### Expected API (期望的 API)
just like `--n-cssvar`
### Examples
https://github.com/TuSimple/naive-ui/pull/1820
https://github.com/TuSimple/naive-ui/pull/1830
### Components
* [x] let CSS var prefixed with n for `alert`
* [x] let CSS var prefixed with n for `anchor `
* [x] let CSS var prefixed with n for `auto-complete`
* [x] let CSS var prefixed with n for `avatar`
* [x] let CSS var prefixed with n for `back-top`
* [x] let CSS var prefixed with n for `badge`
* [x] #1903
* [x] let CSS var prefixed with n for `button`
* [x] #1919
* [x] #1921
* [x] let CSS var prefixed with n for `carousel`
* [x] #1933
* [x] #1935
* [x] #1899
* [x] #1939
* [x] #1941
* [x] #1943
* [x] #1945
* [x] #1954
* [x] #1959
* [x] #1961
* [x] #1962
* [x] #1963
* [x] #1964
* [x] let CSS var prefixed with n for `dynamic-tags`
* [x] #1987
* [x] #1988
* [x] #1996
* [x] #1997
* [x] #1998
* [x] #1989
* [x] #1990
* [x] #1991
* [x] #1983
* [x] #1982
* [x] #1977
* [x] let CSS var prefixed with n for `legacy-grid`
* [x] #1976
* [x] #1975
* [x] let CSS var prefixed with n for `log`
* [x] #1969
* [x] #1968
* [x] #1967
* [x] #1958
* [x] #1957
* [x] #1956
* [x] #1950
* [x] #1949
* [x] #1948
* [x] #1929
* [x] #1926
* [x] #1924
* [x] #1923
* [x] #1918
* [x] let CSS var prefixed with n for `select`
* [x] let CSS var prefixed with n for `select-menu`
* [x] let CSS var prefixed with n for `selection`
* [x] #1914
* [x] #1912
* [x] #1911
* [x] #1908
* [x] let CSS var prefixed with n for `statistic`
* [x] let CSS var prefixed with n for `steps`
* [x] let CSS var prefixed with n for `switch`
* [x] let CSS var prefixed with n for `table`
* [x] let CSS var prefixed with n for `tag`
[...]

PR #1875, opened 2021-12-17
feat(transfer):`transfer` uses `n` as CSS vars prefix

amadeus711:
ref #1842


07akioni:
@amadeus711 需要 rebase 一下最新的，我改了个 bug

PR #1910, opened 2021-12-20
feat(spin): uses `n` as CSS vars prefix

timeTravelCYN:
ref #1842 #1908


07akioni:
@timeTravelCYN Changelog

PR #2101, opened 2022-01-03
Docs: Fix Typo

OrbisK:
Fix Typo. `TabPane Props` is `Tab Props`


OrbisK:
Note:
I am not sure but i think something is wrong in `Tab Props/TabPane Props`. I dont see any usage of the prop `closeable`. I briefly watched the source code, but dont have any time to dig deeper into it.

PR #2125, opened 2022-01-06
docs(modal): add onClose and onMaskClick properties

liuqh0609:
The current version of the document does not have these two attributes, but I feel that they are still quite commonly used so I added them


Talljack:
英文文档一并修改

07akioni:
onClose 算在 Card 和 Dialog 里面了，所以没在 Modal 里面

PR #2144, opened 2022-01-10
fix(radio.cssr.ts): optimize mouse movement style

feligo887:
optimization icon enable: Mouse moving effect is gesture effect, disable effect is disabled

### DIFF
--- a/src/radio/src/styles/radio.cssr.ts
+++ b/src/radio/src/styles/radio.cssr.ts
@@ -40,6 +40,7 @@ export default cB('radio', `
     bottom: 0;
     opacity: 0;
     z-index: 1;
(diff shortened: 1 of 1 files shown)


src/radio/src/styles/radio.cssr.ts
```diff
@@ -115,6 +116,9 @@ export default cB('radio', `
     ]),
     cE('label', {
       color: 'var(--n-text-color-disabled)'
-    })
+    }),
+    cB('radio-input', `
+    cursor: not-allowed;
+  `)
```
07akioni, line 122:
```suggestion
      cursor: not-allowed;
    `)
```

PR #2174, opened 2022-01-13
feat: expose useTheme and lightTheme

jaulz:
It would be great if the library could expose `useTheme` and `lightTheme` so it is easily possible to write custom components the same way as this library does.

### DIFF
--- a/src/composables/index.ts
+++ b/src/composables/index.ts
@@ -1,1 +1,2 @@
 export { useThemeVars } from './use-theme-vars'
+export { useTheme } from './../_mixins/use-theme.ts'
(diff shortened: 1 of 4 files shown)


07akioni:
Are you sure to do this? I can't figure a good scene for it. The built-in theme system is somehow complex for users and may be refactored after some time (although not in short term). Most of time `useThemeVars` is enough.

jaulz:
@07akioni i would like to create a component that looks exactly like the Input component. Would you have any other good idea how to achieve it without copying styles? I wouldn't mind if we prefix the `useTheme` with an underscore to indicate that's it's internal and hence can change.

07akioni:
> @07akioni i would like to create a component that looks exactly like the Input component. Would you have any other good idea how to achieve it without copying styles? I wouldn't mind if we prefix the `useTheme` with an underscore to indicate that's it's internal and hence can change.

If so you would also need create css-render style for it, there would be a lot of work. So why don't you just import from library path like `naive-ui/es/_mixins/xxx`? I think export private API through index is not a good choice.

jaulz:
You are right... I will try it another way first and will close this PR for the time being. Thanks for your input! 😊

PR #2191, opened 2022-01-16
feat: export ThemeCommonVars and make type augmentable

jaulz:
It would be cool to make the theme type augmentable so we can use extend the theme with custom variables `useTheme`. An extension could look like this:
```typescript
import 'naive-ui';

declare module 'naive-ui' {
  export interface ThemeCommonVarsExtension {
      primaryColorActive: string
  }
}
```

### DIFF
--- a/src/composables/use-theme-vars.ts
+++ b/src/composables/use-theme-vars.ts
@@ -2,8 +2,11 @@ import { computed, ComputedRef, inject } from 'vue'
 import { configProviderInjectionKey } from '../config-provider/src/ConfigProvider'
 import { commonLight } from '../_styles/common'
 import type { ThemeCommonVars } from '../_styles/common'
+import type { ThemeCommonVarsExtension } from '../config-provider'
 
-export function useThemeVars (): ComputedRef<ThemeCommonVars> {
+export function useThemeVars (): ComputedRef<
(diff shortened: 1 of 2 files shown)


07akioni:
Have you tried it? I know interface can be augmented but not sure about whether an interface inside package is augmentable. What I know is namespace can be augmented in userland.

I still have a question. How would you use it? Modify vars directly? How can they be consumed?

jaulz:
Yep, I added the changes locally and it is working for me 😊 

First I created a `types/naive-ui.d.ts` file where I augment the interface:
```typescript
import 'naive-ui';

declare module 'naive-ui' {
  export interface ThemeCommonVarsExtension {
      primaryColorActive: string
  }
}
```

Afterwards I can pass the new key as part of the theme to the `ThemeProvider`:
image

And eventually I can use `useThemeVars` with the new key:
image

It just makes things easier because I can use that one theme context without having to add an own context e.g. and the nice thing is that this change is just about types 😊

jaulz:
I wonder why the build is failing... in my environment the build command runs through without issues 🤔

07akioni:
> Details

Me too. Don't worry, I'll fix it. Just ignore it.

jaulz:
Thanks! Let me know if I should add/change anything in the PR itself..

PR #2211, opened 2022-01-18
revert: pr 2162

kungege:
ref [#2162](https://github.com/TuSimple/naive-ui/pull/2162)
image

### DIFF
--- a/src/_internal/selection/src/styles/index.cssr.ts
+++ b/src/_internal/selection/src/styles/index.cssr.ts
@@ -113,7 +113,7 @@ export default c([
       position: relative;
       z-index: auto;
       display: flex;
-      padding: var(--n-padding-single);
+      padding: 3px 26px 0 12px;
       flex-wrap: wrap;
       align-items: center;
       width: 100%;


amadeus711 (approved):

amadeus711 (changes requested):
不应该在这里改，去申明变量的地方修改

kungege:
> 不应该在这里改，去申明变量的地方修改

改变量会影响`.base-selection-input`以及`.base-selection-overlay`的样式

jaulz:
(sorry, in case it was mentioned before but I don't understand Chinese 😊 )

Instead of reverting the change I think it would rather make sense to change this line:
https://github.com/TuSimple/naive-ui/blob/c3b574863b9abefd9765a726b13b4a49fe5ae155/src/_internal/selection/src/styles/index.cssr.ts#L265

to:
```css
padding: 3px 7px 3px 0
```

This would ensure that the style variable is used and is consistent to the rest and it would ensure that there is a proper padding:
image

kungege:
> 

@jaulz  hi 😊, as you see, when the `tag` wraps, the space between the `upper tag` and the `lower tag`  is 6px, I think this is not the result we want
image

jaulz:
I would rather say that this is the correct behaviour because in my opinion each tag should have the height (including padding) of the element itself.

PR #2216, opened 2022-01-19
feat: expose light theme

jaulz:
Expose `lightTheme` so it can be used in a similar way to `darkTheme`. This would be helpful to grab some of the default values while creating the `themeOverrides`. Like we discussed before it would be possible to import it directly from the file but why not export it the same way as we do with `darkTheme`? 😊


jaulz:
Thanks!

PR #2227, opened 2022-01-20
fix(select): tags container loses top padding

nooooooom:


jaulz:
This is more or less a duplicate of https://github.com/TuSimple/naive-ui/pull/2211 but offers also a valid solution.

nooooooom:
> This is more or less a duplicate of #2211 but offers also a valid solution.

Yes, I didn't notice that a similar PR already existed before push. 😋

We decided to work on this branch.

src/_internal/selection/src/styles/index.cssr.ts
```diff
@@ -113,7 +114,9 @@ export default c([
       position: relative;
       z-index: auto;
       display: flex;
-      padding: var(--n-padding-single);
+      padding-top: 3px;
```
jaulz, line 117: If we anyway introduce a new variable, would it be possible to make this themable at the same time?

PR #2232, opened 2022-01-21
docs(table): fix single-column and single-line description

eagleoflqj:
https://github.com/TuSimple/naive-ui/blob/5efe8ae40d52b5fac5137d1bf7f86e1c009697fd/src/table/src/styles/index.cssr.ts#L89-L105

### DIFF
--- a/src/table/demos/enUS/striped.demo.md
+++ b/src/table/demos/enUS/striped.demo.md
@@ -1,4 +1,4 @@
-# 条纹
+# Stripe
 
 ```html
 <n-space vertical>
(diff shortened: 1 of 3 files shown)


src/table/demos/enUS/striped.demo.md
```diff
@@ -18,7 +18,7 @@
         <td>反常的</td>
         <td>彻底废除</td>
```
amadeus711, line 19:
```suggestion
        <td>彻底废除</td>
```
These also need translation

PR #2245, opened 2022-01-23
Update custom-request.demo.md

bruceeewong:
Hi, there is an incorrect usage at line 45: we should pass the "file.file", the binary prop, instead of the `file` param that comes from `customRequest`. Otherwise we will find axios actually sends string "[object, Object]" in the Form Data, not the binary.


Talljack:
Thank you, Please both update file of zhCN folder.

PR #2263, opened 2022-01-24
Feat(dynamic tag) add input props prop

NickyAlfEdge:
Queried on Discord regarding the addition of a max-input-length prop for the NDynamicTag Component's input field. It was suggested that the input-props prop be added to the component to allow for the nested input element to have attributes added to it on creation, for the purposes of better controlling the default input within the NDynamicTag.

I have made the aforementioned changes. I will aim to make a pr to update the documentation to accompany this pr.

Many thanks


Talljack:
The changelog location is wrong, please refer to CONTRIBUTING.md

src/input/src/Input.tsx
```diff
@@ -1019,6 +1018,7 @@ export default defineComponent({
                 readonly={this.readonly as any}
                 autofocus={this.autofocus}
                 size={this.attrSize}
+                {...this.inputProps}
```
Talljack, line 1021: I think props property has higher priority than inputProp like autofocus

NickyAlfEdge:
> The changelog location is wrong, please refer to CONTRIBUTING.md

Have updated the changelogs, though I'm uncertain if it is completely correct. Apologies, I found the contributing instructions a little unclear.

Talljack:
> > The changelog location is wrong, please refer to CONTRIBUTING.md
> 
> Have updated the changelogs, though I'm uncertain if it is completely correct. Apologies, I found the contributing instructions a little unclear.

I think you are right

src/input/src/Input.tsx
07akioni: I don't think inputProps should has higher priority, since it may break default behavior without any warning.

PR #2264, opened 2022-01-25
docs(NDynamicTag): update n dynamic tag docs, adding input props 

NickyAlfEdge:
Accompanying pr for the documentations to be in line with the update of NDynamicTag having the input-props prop added to it's functionality. PR can be found here: [#2263](https://github.com/TuSimple/naive-ui/pull/2263).


Talljack:
Please write in #2263

PR #2276, opened 2022-01-25
feat(Notification)：add top and bottom props

jekip:


07akioni:
container style

added in 8e3474391401b4ebff3f19b1514ddd55a6111ec7

PR #2288, opened 2022-01-26
fix: fix tiny line between arrow and popover

jaulz:
This PR fixes the tiny line between arrows and popovers. Before:
image

After:
image


07akioni:
I've tried it before. In my memory, antd doesn't have such problem. Maybe we can see how antd do it.

jaulz:
I think they do something similar but just on a JavaScript level:
image

They also have a tiny difference:
image

07akioni:
> I think they do something similar but just on a JavaScript level: image
> 
> They also have a tiny difference: image

Can we apply their solution

jaulz:
Having an even closer look:
image

I think distance is clear but the second part is unclear for me:
image

(see https://github.com/ant-design/ant-design/blob/master/components/tooltip/style/index.less)

07akioni:
The problem is fixed by using opaque background.

PR #2329, opened 2022-02-02
feat: add `defaultValidationStatus` prop to `Form`

jaulz:
This PR will add a new `defaultValidationStatus` prop to `Form` which for example let's you change the validation status to warnings. In my case I would like to show client based validations as warnings and if the user anyway clicks on submit it should be displayed as errors.

### DIFF
--- a/CHANGELOG.en-US.md
+++ b/CHANGELOG.en-US.md
@@ -19,6 +19,7 @@
 - `n-time-picker` adds `placement` prop.
 - `n-tree-select` adds `placement` prop.
 - `n-card` adds `header-extra-style` prop.
+- `n-form` adds `default-validation-status` prop.
 
 ### Fixes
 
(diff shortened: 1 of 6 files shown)


07akioni:
I think the prop may be useful in some case. However I don't undestand your scene. If users don't trigger validation, how could you know field should be in warning status?

jaulz:
Client validations will be triggered by change event and server validations by submit. I am currently rethinking this scenario but this prop might anyway be helpful to avoid hard coded values 😊

07akioni:
> Client validations will be triggered by change event and server validations by submit. I am currently rethinking this scenario but this prop might anyway be helpful to avoid hard coded values 😊

A default warning status means it is warning by default. Why user may need it?

jaulz:
Okay, let's close it then for the time being 😉

PR #2334, opened 2022-02-03
fix: add proper tabindex when filterable

jaulz:
Currently, when a `Select` is filterable you cannot reach the filter via tab control. This PR will add a proper `tabindex` to the inputs.


07akioni:
You can open select by pressing enter if it's focused. 

I make lazy focus intentionally. Is the behavior problematic?

07akioni:
Can you rebase it? I want to see vercel preview.

jaulz:
@07akioni here you go 😊

07akioni:
> Preview

Normal select's tab focusing is broken.

jaulz:
Hm, what's not working for you? I can still focus the tab and press enter to open it.

07akioni:
Tab focus this

image

jaulz:
@07akioni sorry for the issue I missed that before. I think now it works fine and whenever a select is filterable you can tab properly and can start immediately typing.

07akioni:
After esc is pressed in filterable select, the select box should still be focused.

image

I think maybe provide a prop to let user choose whether to focus input immediately is better.

jaulz:
Okay, sure. Any suggestion for the name of the prop? 😊

07akioni:
> Okay, sure. Any suggestion for the name of the prop? 😊

Maybe `focus-input-when-filterable`? Not quite sure.

Issue #1293, opened 2021-10-08
Add hold event to change value of <n-input-number/>


sondh0127:
### This function solves the problem (这个功能解决的问题)
we can change the value of input number in single hold click
### Expected API (期望的 API)
The element-plus team use this 'v-repeat-click' directive
```
 <span
  v-if="controls"
  v-repeat-click="increase"
  class="el-input-number__increase"
  role="button"
  :class="{ 'is-disabled': maxDisabled }"
  @keydown.enter="increase"
>
  <i :class="`el-icon-${controlsAtRight ? 'arrow-up' : 'plus'}`"></i>
</span>
```
https://github.com/element-plus/element-plus/blob/dev/packages/directives/repeat-click/index.ts

Talljack:
Which demo in element?

sondh0127:
> Which demo in element?
This one
https://element-plus.org/en-US/component/input-number.html

Talljack:
> > Which demo in element?
> > This one
> > https://element-plus.org/en-US/component/input-number.html

I think you can be specific like https://element-plus.org/en-US/component/input-number.html#steps

PR #2391, opened 2022-02-13
feat(n-input-number): Add hold event to change value of <n-input-numb…

NickyAlfEdge:
Functionality for hold click increment/decrement of the n-input-number icons, resolves: [#1293](https://github.com/TuSimple/naive-ui/issues/1293)

Have updated the test cases to accommodate for the changes, but I'm not overly sure if this is an ideal solution. It operates similarly to the approach implemented in Element Plus, albeit the existing implementation doesn't have an onleave event like Element Plus, so if the buttons are held down and then the mouse is moved away from them, then the number continues to increase/decrease (Though the n-input-number component also stays highlighted until it is unfocused anyway). In line with this, I clear the increase/decrease intervals onblur, this keeps in theme with how the inputs are in focus until other elements are focused.


src/input-number/src/InputNumber.tsx
```diff
@@ -457,12 +496,28 @@ export default defineComponent({
...
+                                ? undefined
+                                : this.handleMinusClick
+                            }
+                            onMouseup={
+                              !this.minusable ||
+                              this.mergedDisabled ||
+                              this.readonly
+                                ? undefined
+                                : this.clearMinusStepInterval
+                            }
```
07akioni, line 519: This event should be bound to button, not icon

07akioni:
Increment is triggered even after mouse left button and up.

07akioni:
Increment should has a trigger threshold. For example only if you hold button after 1s, we start to increase the value.

NickyAlfEdge:
Have resolved the issues you mentioned.

- Added onMousedown, onMouseup and onMouseleave events to the NButton for use with this functionality.
- Moved the events to the parent button and not the icon.
- Added a 1 second delay prior to any incrementing/decrementing of the input value.
- Updated the test cases to reflect these changes.

Thanks so much for the comments, I was unsure whether to add anymore events to the existing Button. Anyway, I hope the latest changes will suffice, I tried to abide by the general approaches within the codebase.

NickyAlfEdge:
Never-mind, I've identified an issue. It doesn't seem to interact well with Min and Max props, will keep trying to figure this out.

NickyAlfEdge:
Apologies, the min / max prop issue should now be resolved. Any issues then please do highlight them and I will do my best to amend them.

07akioni:
I've refactor your commit. You can have a look at what should be changed in your PR.

758eed5d1ef6cfb0c6922c346c9e9908a86f09ad

PR #2508, opened 2022-02-24
feat: add color prop to Typography

jaulz:
This PR will add a `color` prop to the `Typography` components to support custom colors.

### DIFF
--- a/src/typography/demos/enUS/header.demo.vue
+++ b/src/typography/demos/enUS/header.demo.vue
@@ -52,6 +52,9 @@ Make your headers pop with a little bar to the left. Much less boring.
       </n-text>
     </n-h1>
     <n-h2>cos(x)</n-h2>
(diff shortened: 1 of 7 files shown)


07akioni:
I prefer style if there's no nested DOM.

Add color may add api complexity.

jaulz:
Hm, I just thought to make the API consistent and do not let the developer rely on setting a CSS variable whose name could change later. Having a dedicated prop would definitely make it future proof.

07akioni:
For components which already had `color` prop, most of them can't be directly styled by inline style, for example `button`, `progress`.

I think if the prop is not necessary, we prefer not adding it.

For example, long time ago, popover has `width` and `max-width` prop. However I found it can be achieved easily by `style`, so I removed it later.

jaulz:
Okay, then I will close it for the time being 😊

PR #2527, opened 2022-03-01
docs(config-provider): inline-theme-disabled description

07akioni:

### DIFF
--- a/src/config-provider/demos/enUS/inline-theme-disabled.demo.vue
+++ b/src/config-provider/demos/enUS/inline-theme-disabled.demo.vue
@@ -3,7 +3,7 @@
 
 By default, naive-ui uses inline style to hold theme variables, so every component is bound with large amount of inline CSS. If you need SSR, or just make your devtools clear, you can enable `inline-theme-disabled` prop.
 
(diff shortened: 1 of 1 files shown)


src/config-provider/demos/enUS/inline-theme-disabled.demo.vue
```diff
@@ -3,7 +3,7 @@
 
 By default, naive-ui uses inline style to hold theme variables, so every component is bound with large amount of inline CSS. If you need SSR, or just make your devtools clear, you can enable `inline-theme-disabled` prop.
 
-Note that if you will change `theme-overrides` frequently, it's not recommended to enable the prop. It will generate many useless style element during `theme-overrides` changes.
+Note that if you will change `theme-overrides` frequently, it's not recommended to enable the prop. Many useless style elements will be keeped after `theme-overrides` changes.
```
07akioni, line 6:
```suggestion
Note that if you will change `theme-overrides` frequently, it's not recommended to enable the prop. Many outdated style elements will be keeped after `theme-overrides` changes.
```

PR #2555, opened 2022-03-04
fix(time): does not provide `getTimezoneOffset`

Sepush:

ref #2545 

maybe the same case with https://github.com/marnusw/date-fns-tz/issues/133 it fixed in `date-fns-tz 1.3.0`


Baiyuetribe:
Nuxt3也有该错误，我还以为是特殊情况。感谢修复

phcent:
依然报这个错误哦 https://codesandbox.io/s/woeq78

PR #2599, opened 2022-03-09
Fix: use full paths with .js extension on esm mode

rhengles:
As explained [on Discord](https://discord.com/channels/842427094214770768/842427094214770771/950748048080916561):

When using `vite@2.8.6` and `vite-ssg@0.17.11`, the SSR build fails with these kind of errors:
```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/{pnpm path}/node_modules/naive-ui/es/alert/src/Alert' imported from /{pnpm path}/node_modules/naive-ui/es/alert/index.js
```
This is not the first error, I've modified many places inside the compiled naiveui package, like this:
```diff
 // example
 // When dealing with files, vite-ssg only finds them if I add the extension:
-export { default as dateNbNO } from './date/nbNO';
+export { default as dateNbNO } from './date/nbNO.js';
 // When dealing with directories, I must add the index file path:
-export { createLocale } from './utils';
+export { createLocale } from './utils/index.js';
```
Related:
https://github.com/antfu/vite-ssg/issues/201
https://github.com/antfu/vite-ssg/issues/205

This PR add a post-build step with Babel to automatically resolve all relative file imports, adding the `.js` extension or `/index.js` when importing from a directory. (obs: it's also necessary to change the locale import from date-fns to `date-fns/locales/en-US/index.js`)

As discussed on discord, I don't expect this PR to be merged, but would like to have a solution to be able to use the latest vite-ssg. Thank you very much.

### DIFF
--- a/babel.config.es.js
+++ b/babel.config.es.js
@@ -1,0 +1,10 @@
+module.exports = {
+  plugins: [
+    [
+      'babel-plugin-custom-import-path-transform',
+      {
+        transformImportPath: './scripts/transform-es-import-path.js'
+      }
+    ]
+  ]
+}
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "scripts": {
     "start": "pnpm run dev",
     "dev": "pnpm run clean && pnpm run gen-version && pnpm run gen-volar-dts && cross-env NODE_ENV=development vite",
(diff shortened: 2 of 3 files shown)


rhengles:
This is not the only package I had to modify to get the build running - I also applied this solution to `vueuc` and `css-render`.

rhengles:
Related: https://github.com/antfu/vite-ssg/discussions/217

PR #2647, opened 2022-03-16
feat(input): add ability to custom password toggle icon

izcream:
This first time to make PR to this projects. Sorry if I do anything wrong.

ps. you can change props name to another that you thinks it meaningful😊

### DIFF
--- a/src/input/src/Input.tsx
+++ b/src/input/src/Input.tsx
@@ -15,7 +15,8 @@ import {
   provide,
   InputHTMLAttributes,
   TextareaHTMLAttributes,
-  Fragment
+  Fragment,
+  Component
 } from 'vue'
 import { useMergedState, useMemo } from 'vooks'
 import { getPadding } from 'seemly'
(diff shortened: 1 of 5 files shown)


DannyFeliz:
It would be cool it you could add slots for these icon 👀

izcream:
That's on my plan about slot implementation but still newbie on this repo architecture, need a bit time to understand this😆😍

src/input/src/Input.tsx
```diff
@@ -1107,6 +1110,16 @@ export default defineComponent({
           )}
           {!this.pair &&
             resolveWrappedSlot($slots.suffix, (children) => {
+              const passwordVisibleIcon = this.passwordVisibleIcon ? (
+                h(this.passwordVisibleIcon)
```
07akioni, line 1114: I think make it a slot is a better way. For example

```
// from src/_utils
resolveSlot(this.$slots.['password-visible'], () => [<EyeIcon />])
```

07akioni: Since it's more convient to compose.

izcream: awesome😊

PR #2665, opened 2022-03-22
`n-data-table` adds generic to type `DataTableSortState`

OrbisK:


OrbisK:
not ready. Missed DataTableSortState  :)

07akioni:
I'm not sure why this is needed?

The problem won't be solved by the generic.

For example, if you are using it in callback

```
setup () {
  return {
    onUpdateSortState(s: DataTableSortState<X>) {}
  }
}
```

`onUpdateSortState` won't be acceptable to `n-data-table` since data-table accepts `DataTableSortState<X>`.

A type assertion can't be omitted. It's just transferred from one place to another place.

OrbisK:
you are right. i didnt thougt that far ahead. Unlucky 🥲

PR #2691, opened 2022-03-29
Document's table meta row sets wrong number of columns

ArtemisDicoTiar:
The content of the table has only 5 columns but the meta row (the row right below the header. It includes such as "| -- |") had 6 columns.


amadeus711:
Hello, there is no problem with your PR, but the branch is wrong. It should be `checkout` from docs.

ArtemisDicoTiar:
@amadeus711 Thanks. I will close this one and re-PR with proper checkout :)

PR #2723, opened 2022-04-01
🐛 fix(input-number): add `justify-icon` props

izcream:
justify-icon props can make input-number control(-+) icon stay around input like image below

image

you can change props name to make-sense more than this(I have no idea what this should name😂)


07akioni:
Looks somehow reasonable. I would dig into it this weekend.

src/input-number/src/InputNumber.tsx
```diff
@@ -562,7 +566,45 @@ export default defineComponent({
...
+                              aria-disabled={true}
+                            >
+                              {{
+                                default: () => <RemoveIcon />
+                              }}
+                            </NBaseIcon>
+                          )
+                        }}
+                      </NxButton>
+                    ),
```
07akioni, line 607:
```suggestion
                  </NxButton>,
```

src/input-number/src/InputNumber.tsx
```diff
@@ -562,7 +566,7 @@ export default defineComponent({
                             </NBaseIcon>
                           ])
                       }}
-                    </NxButton>,
+                  </NxButton>,
```
07akioni, line 569:
```suggestion
                   </NxButton>,
```

PR #2809, opened 2022-04-18
[Untested] Simple directory support

### LINKED ISSUE #2765: Improved multiple file upload handling
When dropping multiple files into the upload area, we can obtain the full path of files within sub-directories. This is useful if the user drops one or more folders, instead of individual files. It is made possible by using non-standard APIs that (I believe) are supported by major browsers (https://caniuse.com/?search=webkitGetAsEntry). As a result, you get a full path like `my-folder/images/foo.jpeg`

Here is the code to support this behaviour.

```
async function drop(e) {
  e.preventDefault()
  e.stopPropagation()
  let transfer = e.dataTransfer
  let entries = await getAllFileEntries(transfer.items)

  entries.forEach(async entry => {
    entry.file(async file => {
      console.log(entry.fullPath, file)
    })
  })
}

async function getAllFileEntries(dataTransferItemList) {
  let fileEntries = []
  let queue = []
  for (let i = 0; i < dataTransferItemList.length; i  ) {
    queue.push(dataTransferItemList[i].webkitGetAsEntry())
  }

  while (queue.length > 0) {
    let entry = queue.shift()
    if (entry.isFile) {
      fileEntries.push(entry)
    } else if (entry.isDirectory) {
      let reader = entry.createReader()
      queue.push(...await readAllDirectoryEntries(reader))
    }
  }
  return fileEntries
}

async function readAllDirectoryEntries(directoryReader) {
  let entries = []
  let readEntries = await new Promise(directoryReader.readEntries)
  while (readEntries.length > 0) {
    entries.push(...readEntries)
    readEntries = await new Promise(directoryReader.readEntries)
  }
  return entries
}
```
### Expected API (期望的 API)
The API can remain unchanged, but `UploadFileInfo` should have `fullPath` or similar.

lwansbrough:
Resolves #2765


lwansbrough:
Because `webkitdirectory` supports directories more simply than the other proposed method in #2765 I have used it instead of the more complicated approach.

PR #2958, opened 2022-05-17
Fix calendar today panel change

kevincoenegrachts:
The calendar on-panel-change returned the wrong month object when clicking today.
image

It returned the month timestamp instead.


package.json
```diff
@@ -7,7 +7,7 @@
   "types": "es/index.d.ts",
   "scripts": {
     "start": "pnpm run dev",
-    "dev": "pnpm run clean && pnpm run gen-version && pnpm run gen-volar-dts && cross-env NODE_ENV=development vite",
+    "dev": "pnpm run clean && pnpm run gen-version && cross-env NODE_ENV=development vite",
```
07akioni, line 10:
```suggestion
    "dev": "pnpm run clean && pnpm run gen-version && pnpm run gen-volar-dts && cross-env NODE_ENV=development vite",
```

PR #2972, opened 2022-05-19
fix: export the CountdownTimeInfo interface

DannyFeliz:

### DIFF
--- a/src/countdown/src/Countdown.tsx
+++ b/src/countdown/src/Countdown.tsx
@@ -9,7 +9,7 @@ import {
 } from 'vue'
 import type { ExtractPublicPropTypes } from '../../_utils'
 
-interface TimeInfo {
+export interface TimeInfo {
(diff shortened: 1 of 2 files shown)


src/countdown/src/Countdown.tsx
```diff
@@ -9,7 +9,7 @@ import {
 } from 'vue'
 import type { ExtractPublicPropTypes } from '../../_utils'
 
-interface TimeInfo {
+export interface TimeInfo {
```
07akioni, line 12: Rename it to `CountdownTimeInfo` and export it from index of Countdown component.

07akioni: For exposed interface we need to give it a scope.

DannyFeliz: That makes sense, thank you.

PR #2987, opened 2022-05-22
fix: menu-item--selected::before not working

1161906592:


1161906592:
lQLPDhtyHmt7Mh_NAq7NAvmwhNXYfiiuU94Ci6U-tcDSAA_761_686
menu-item--selected::before 应该是 menu-item-content--selected::before

PR #3003, opened 2022-05-23
bug: expand upload trigger to full upload box

gimler:
The n-upload div has `width: 100%` but not the n-upload-trigger div.

before
image

after
image


07akioni:
I've digged into the problem. It's a regression after a refactor before.

The key pointer is here (after changed).

image

100% wasn't applied to correctly class before.

PR #3043, opened 2022-06-02
feat: add expandOnClick props in n-tree

Yhspehy:
Add expandOnClick props in n-tree. closes [#2949](https://github.com/TuSimple/naive-ui/issues/2949).

### DIFF
--- a/src/tree/src/Tree.tsx
+++ b/src/tree/src/Tree.tsx
@@ -140,6 +140,10 @@ const treeProps = {
     type: Boolean,
     default: true
   },
+  expandOnClick: {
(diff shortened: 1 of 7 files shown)


src/tree/src/Tree.tsx
```diff
@@ -140,6 +140,10 @@ const treeProps = {
     type: Boolean,
     default: true
   },
+  expandOnClick: {
+    type: Boolean,
+    default: false
+  },
```
07akioni, line 146:
```
  expandOnClick: Boolean,
```

PR #3074, opened 2022-06-09
fix: set min height of alert

jaulz:
The alert component should have at least the height of the icon. Right now, the icon overlapses the container in case it's height is higher than the container itself.

### DIFF
--- a/src/alert/src/styles/index.cssr.ts
+++ b/src/alert/src/styles/index.cssr.ts
@@ -33,6 +33,7 @@ export default cB('alert', `
   background-color: var(--n-color);
   text-align: start;
   word-break: break-word;
+  min-height: var(--n-icon-size);
 `, [
   cM('closable', [
     cB('alert-body', [


XieZongChen (approved):

07akioni:
The solution may not fix the problem.

image

If you want empty alert works, I think you can pass this to `n-alert`.

```
<template #header>&nbsp;</template>
```

jaulz:
@07akioni my problem is the other way around: I want bigger icons but since the icon is absolutely positioned the box doesn't take the icon's size into consideration when calculating the layout.

07akioni:
> @07akioni my problem is the other way around: I want bigger icons but since the icon is absolutely positioned the box doesn't take the icon's size into consideration when calculating the layout.

The design is intended to solve customization problem. Icon's position is customizable, so I used absolute position for icon.

If we want to make it take space we may need to figure out a way to keep icon's position customizable.

jaulz:
Hm, how's the icon's position customizable right now? Based on the code I can't see any possibility to move it anywhere else except for using specific custom styles.

07akioni:
> Hm, how's the icon's position customizable right now? Based on the code I can't see any possibility to move it anywhere else except for using specific custom styles.

`iconMargin` can be customized.

jaulz:
If it's just about the margin I still don't understand why position `absolute` is used. A simple flex layout should do the same and even respect the size of the icon.

07akioni:
> If it's just about the margin I still don't understand why position `absolute` is used. A simple flex layout should do the same and even respect the size of the icon.

Margin can make icon placed in arbitrary position.

Some designers in companys may request it placed in some strange position.

07akioni:
However I think this is a problem, it should be fixed in some way.

PR #3120, opened 2022-06-17
fix(tag): missing bordered attribute in demo

lrstanley:
Fixed `bordered` reference for the tag demo. Not sure if that requires a changelog update.


07akioni:
Docs update doesn't require update

PR #3126, opened 2022-06-17
feat: add a new prop(adjustable) so that the drawer can adjust its width/height

Mr-ccyou:


07akioni:
Can you add a demo?

Mr-ccyou:
> Can you add a demo?

ok, i will.

PR #3234, opened 2022-07-05
feat(drawer): add a new prop(adjustable) so that the drawer can adjust its width/height

Mr-ccyou:


Mr-ccyou:
这个n-tree测试用例没跑过,
image

PR #3259, opened 2022-07-09
feat(scrollbar): rtl

aminmokhtari94:


07akioni:
I've a question. Does scrollbar need a RTL version?

I checked a few RTL website and find their (native browser) scrollbar is on the right side. I'm not sure if it's caused by system settings (it's using a LTR locale).

If browser doesn't support a RTL scrollbar I think it may not neccessary to make scrollbar component RTL.

If browser supports it could you please provide some website link or snapshot?

PR #3275, opened 2022-07-12
feat(message): rtl

aminmokhtari94:


leila01010:
> 

You can add new variable like `--n-icon-margin-rtl` in `message/styles/_common.ts` file and then add it in `Message.tsx` file.
You can look at the `pagination` component, I did it

aminmokhtari94:
@07akioni @amadeus711
Is there a problem?

07akioni:
> @07akioni @amadeus711 Is there a problem?

I don't think this is a proper solution. Since it's behavior is implicit.

PR #3383, opened 2022-07-28
docs(dialog): move style to DialogReactive props (en)

OrbisK:
Missmatch between en and cn version. I assume that en is wrong.


OrbisK:
missed that in last PR 😅

PR #3385, opened 2022-07-29
feat(button): add button useAsync props

QAQandBBQ:


07akioni:
I think this should be a composable, not a prop.

PR #3434, opened 2022-08-06
chore: from jest to vitest

Sepush:

jest
naive-jest

vitest

vitest-naive

reduce to 6/10 of before

only thing left is there are a few test case use the magic if `setTimeout` we should not write test that way so I temporarily commented out to get test pass I leave TODO in the codebase.
I will try to find out a way to solve them, and make a tracking issue to list and track them.

waiting for https://github.com/vitest-dev/vitest/issues/1191


Sepush:
still get issue on ci but all test pass on my computer with node 18.7.0
I think there still bug

Sepush:
是 drawer 漏掉了 一个 setTimeout

PR #3470, opened 2022-08-09
fix(n-input): duplicate placeholder if disabled

Lucky-Ya-Q:
fix https://github.com/TuSimple/naive-ui/issues/3467


07akioni:
We can't simply remove it since they are introduced to fix a safari only style bug.

07akioni:
I've fixed it.

PR #3489, opened 2022-08-11
types(n-input-number): update type `Size` to allow `tiny`

OrbisK:

### DIFF
--- a/src/input-number/demos/enUS/size.demo.vue
+++ b/src/input-number/demos/enUS/size.demo.vue
@@ -1,11 +1,10 @@
 <markdown>
 # Size
-
-`small`, `medium`, `large`.
 </markdown>
 
 <template>
   <n-space vertical>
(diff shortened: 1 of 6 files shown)


src/input-number/demos/enUS/size.demo.vue
```diff
@@ -1,11 +1,10 @@
 <markdown>
 # Size
-
-`small`, `medium`, `large`.
```
OrbisK, line 4: i have omitted this to remain consistent with `n-input`

src/input-number/src/interface.ts
```diff
@@ -1,4 +1,6 @@
 export type OnUpdateValue = (value: number | null) => void
+export type Size = 'tiny' | 'small' | 'medium' | 'large'
```
OrbisK, line 2: not sure if this should better import `Size` from `n-input`

PR #3525, opened 2022-08-16
docs(timeline): set value to `right` in item-placement demo

Jiangmenghao:
# timeline

- set value to `right` in `item-placement` demo (consistent with the Chinese version)

---

# 时间线

- `item-placement` demo 的值设为`right`（与中文文档保持一致）

### DIFF
--- a/src/timeline/demos/enUS/item-placement.demo.vue
+++ b/src/timeline/demos/enUS/item-placement.demo.vue
@@ -5,7 +5,7 @@ Left. .thgiR
 </markdown>
 
 <template>
-  <n-timeline item-placement="left">
+  <n-timeline item-placement="right">
     <n-timeline-item content="Oops" />
     <n-timeline-item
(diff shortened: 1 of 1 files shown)


src/timeline/demos/enUS/item-placement.demo.vue
```diff
@@ -1,11 +1,11 @@
 <markdown>
 # Item placement
 
-Left. .thgiR
```
amadeus711, line 4: This is no problem.

amadeus711 (changes requested):

src/timeline/demos/enUS/item-placement.demo.vue
Jiangmenghao: Oh ~ I now understand the humor 🤣

amadeus711 (approved):

PR #3612, opened 2022-08-28
docs(data-table): adjust border.demo.vue

OrbisK:
border was a bit confusing 🥲 

- [ ] cn translation needed


src/data-table/demos/zhCN/border.demo.vue
```diff
@@ -1,5 +1,5 @@
 <markdown>
-# 无边框 & 单行
+# Unbordered & bordered & all between
```
07akioni, line 2:
```suggestion
# 无边框 & 有边框 & 其他
```

PR #3633, opened 2022-08-30
docs(select): placeholder default is `undefined`

OrbisK:
placeholder should default to `undefined`. Like input docs.

whould be confusing for other languages.


amadeus711:
image
There is a problem with this PR.

OrbisK:
Maybe we should link the default to i18n. So we don't specify the default as kinda hard coded.

07akioni:
added i18n in 09be4975920945817842ecebc13104cbfdf5956f

PR #3654, opened 2022-09-01
chore: bump ts to 4.8

Sepush:

TypeScript(from 4.7) now enforces stricter checks that the given type is actually an object. As a result, values with the types unknown and never (and more rarely, just bare null and undefined) can no longer be spread into JSX elements.
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#stricter-spread-checks-in-jsx

Unconstrained Generics No Longer Assignable to {}
https://devblogs.microsoft.com/typescript/announcing-typescript-4-8-beta/#unconstrained-generics-no-longer-assignable-to


Sepush:
4.8 is available but have some issue(have been fixed but did not realse) https://github.com/vuejs/core/issues/6554

amadeus711 (approved):

PR #3691, opened 2022-09-07
docs: improve Web Types generation; include available documentation

### LINKED ISSUE #3675: Provide information about slots in web-types or *.d.ts files
When developing projects with Naive UI in WebStorm I get lots of false positive warnings about unrecognized slot names, which is pretty annoying. It seems information about the slots needs to be provided either via web-types or *.d.ts definitions (see this [issue](https://youtrack.jetbrains.com/issue/WEB-55215#focus=Comments-27-6426056.0-0)). Could you provide these? It would help a lot, thank you!
### Expected API (期望的 API)
Provide information about slots in type definitions.

piotrtomiak:
This PR aims at improving user experience in WebStorm and other JetBrains IDEs. The generated Web Types will now include available documentation and information about slots. Source navigation and links to online documentation are also added.

Here is a screencast of how code completion works with new Web Types:
 
naive-ui-screencast 2022-09-07 12_36_53

Please note that some of the features as code completion within slot destructing parameters will be available on in WebStorm 2022.3.

WebStorm works with `.d.ts` files, but the experience is worse than with properly created Web Types. Further improvements to integration between `*.d.ts` and Web Types is expected to come in 2022.3.

close #3675

### DIFF
--- a/scripts/post-build/gen-web-types.js
+++ b/scripts/post-build/gen-web-types.js
@@ -3,11 +3,13 @@ const fs = require('fs')
 const path = require('path')
 const { kebabCase } = require('lodash')
 
+const baseDir = path.resolve(__dirname, '../..')
+
 exports.genWebTypes = function genWebTypes () {
   const components = require('../../lib/components')
   const { default: version } = require('../../lib/version')
 
-  const elements = []
+  const vueComponents = []
 
   const scaffold = {
     $schema:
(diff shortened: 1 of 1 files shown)


amadeus711 (approved):

lpj145:
@07akioni we really need this implementation ❤️

PR #3756, opened 2022-09-22
docs: update xicons domain

1161906592:


OrbisK:
i think you have to open this agains docs, not main

PR #3816, opened 2022-10-04
fix dynamic-input wrong index on create click

hedwiggggg:
When the Create button is pressed on DynamicInput, the onCreate handler is called with the wrong index.

### DIFF
--- a/src/dynamic-input/src/DynamicInput.tsx
+++ b/src/dynamic-input/src/DynamicInput.tsx
@@ -185,7 +185,7 @@ export default defineComponent({
       doUpdateValue(newValue)
     }
     function handleCreateClick (): void {
-      createItem(0)
+      createItem(-1)
     }
     function createItem (index: number): void {
       const { value: mergedValue } = mergedValueRef


07akioni:
Is this the expected behavior? When I click at first item's plus button, I mean to add an item after it.

hedwiggggg:
In my opinion, it would be expected that the first time an item is created, it is added at index 0 and the onCreate handler is called with index 0. However, since the index within the createItem is calculated plus one, the index is incorrectly assumed to be 1 during initial creation.

So if you click on create, the onCreate handler is called with index 1 and if you add another item the onCreate handler is called again with index 1. Correct would be 0 and 1.  

I created a stackblitz to demonstrate the issue: https://stackblitz.com/edit/vue-r4fkjn?file=src/App.vue

hedwiggggg:
Again for clarification. I am not interested in whether the index is now 0 or 1. The problem I see is the inconsistency:

```
[click create]: onCreate is called with index 1.
[click add]:    onCreate is called again with index 1
[click add]:    onCreate is called with index 2
```

but it should be called with either 1, 2, 3 or 0, 1, 2 as indexes.

PR #4043, opened 2022-11-16
fix(tree): use `on-load` will cause infinite loop

### LINKED ISSUE #4038: [tree component] use onLoad, an infinite loop will occur if the isLeaf or children has not changed.
2.33.5

### Vue version (Vue 版本)
3.2.45

Chrome(107.0.5304.107)

window_10

### Reappearance link (重现链接)
https://github.com/amihhs/navie_tree_issue

### Reappearance steps (重现步骤)
1. pnpm i 
2. click tree node
3. see browser log

### Expected results (期望的结果)
only load once

### Actual results (实际的结果)
loop load

### Remarks (补充说明)
The situation will happen in a moment：
1. 
```js
async function loadHandle(node) {
    return false
}
```
2. 
```js 
async function loadHandle(node) {
   try{
    ...
    }
  catch(error){
    // Handle errors but do not continue to expose errors
   }
}
```

jizai1125:
closes #4038

### DIFF
--- a/CHANGELOG.en-US.md
+++ b/CHANGELOG.en-US.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## NEXT_VERSION
+
+### Fixes
+
+- Fix `n-tree` use `on-load` will cause infinite loop when not add `children` into parent tree node，closes [#4038](https://github.com/tusen-ai/naive-ui/issues/4038).
+
 ## 2.34.2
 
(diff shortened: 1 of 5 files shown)


07akioni:
fixed in https://github.com/tusen-ai/naive-ui/commit/b74d62a5e9732ee66d38523f6d57bf2a7105e328

PR #4071, opened 2022-11-22
docs(color-picker): modify action props description

jahnli:
docs(color-picker): modify action props description


07akioni:
fixed in beb18e2c4244f8dff8e64ea795209f412ad5bff5

PR #4074, opened 2022-11-22
fix(date-picker): Fix current year can't be selected when some date are disabled

g-plane:


Sepush (approved):

sky124380729:
什么时候发布呀

g-plane:
@07akioni Can you review and merge PR? Thanks.

g-plane:
I found that my solution has bug. Closed.

PR #4113, opened 2022-11-30
fix(anchor):  demo not work

Yhspehy:
#4107


amadeus711:
I verified this PR, so that the modification can make the `n-anchor` demo work normally, but it will cause problems in other official website jumps, so the modification of this problem is subject to #4117.

Yhspehy:
it will not cause problems because we dont't have same filename with demo.md and demo.vue.

PR #4295, opened 2023-01-04
feat: tabs placement props support

faga295:


faga295:
unresolved problems:
* the width of nav in vertical segment type tabs, currently I set the width to 30% of the overall
more problems is going to find...

HyattXt:
什么时候能合并呢，在等这个功能

PR #4341, opened 2023-01-11
Specify createDiscreteApi return type

lbagic-elude:
Narrow down `createDiscreteApi` return type based on provided types.

Currently `createDiscreteApi(['dialog'])` return type includes the following properties `app | dialog | loadingBar | message | notification | unmount`. The return type with proposed change would have just `app | dialog | unmount`

P.S. Sorry for not adding changelog, editing via github :s


lbagic-elude:
Merged/closed with https://github.com/tusen-ai/naive-ui/commit/acf8773d92d0cf297c49e288805c42eeb85a0637

Issue #4420, opened 2023-01-30
Highlight selected/expanded row


shaniqwa:
### This function solves the problem (这个功能解决的问题)
Hi there, so far this library is awesome! 
I would like to know if there is a way to simply set a class to rows when they are selected or in expanded ? (the expanded row   the newly added row)

This is a very common use case where you might want to highlight a row for any reason. of course it makes the UX better when you can highlight what you want to focus on (similar to when you use stripes mode for a table, makes it more readable)

Thanks a lot
Shani
### Expected API (期望的 API)
set a class to  expanded <tr>´expanded’. 
it would be great to control it with the theme config provider. for example add a these properties to the DataTable: "selectedRowColor", "expandedRowColor", "expandedRowContentColor"

shaniqwa:
Opened a PR adding a class to expanded rows: https://github.com/tusen-ai/naive-ui/pull/4426

jahnli:
This issue does not have any recent activity. If you are still experiencing similar problems, open a new error, including a minimal copy of the problem

PR #4426, opened 2023-01-31
feat(data-table):add class to expanded rows to allow styling them

shaniqwa:
- `n-data-table` adds `n-data-table-tr--expanded` class to expanded rows, and `n-data-table-tr n-data-table-tr--expand` class to the additional row. 
 
Partially closing [#4420](https://github.com/tusen-ai/naive-ui/issues/4420), since I need help to do the same for selected rows


src/data-table/src/TableParts/Body.tsx
```diff
@@ -675,7 +675,7 @@ export default defineComponent({
                 } = rowInfo
                 return (
                   <tr
-                    class={`${mergedClsPrefix}-data-table-tr`}
+                    class={`${mergedClsPrefix}-data-table-tr ${mergedClsPrefix}-data-table-tr--expand`}
```
07akioni, line 678:
```suggestion
                    class={`${mergedClsPrefix}-data-table-tr ${mergedClsPrefix}-data-table-tr--expanded`}
```

PR #4428, opened 2023-02-01
fix(steps): number style error when two digits

amadeus711:


amadeus711:
Don't merge. There is a problem with the animation. I will solve it.

PR #4435, opened 2023-02-03
fix(n-select): dropdown options support size `tiny`

### LINKED ISSUE #4398: Select doesn't support size prop
2.34.3

### Vue version (Vue 版本)
3.2.45

Версия 109.0.5414.87

MacOS (13.1 (22C65))

### Reappearance link (重现链接)
https://codesandbox.io/s/interesting-feather-g1vmty

### Reappearance steps (重现步骤)
If we change the default tiny, small and large size theme values, only the font size in the field and its height will change, but it doesn't work with padding like for input, or with font size in the select dropdown.

### Expected results (期望的结果)
It will be great if theme value will be apply for all form components to be consistent when it used in one form with one size.

### Actual results (实际的结果)
Now select look different in form

jizai1125:
closes #4398


ksddesigner:
Thank you! But select don't support padding for sizes.
image
image

jizai1125:
> 

 I keep consistent with other sizes of select because they also don't support.

ksddesigner:
Maybe it will be possible iа add variables for Select in the Theme Editor to users can make it consistent with Input themself?

jizai1125:
> Maybe it will be possible iа add variables for Select in the Theme Editor to users can make it consistent with Input themself?

Yes, it is supported. You can use the themeOverrides prop to override the padding, for example, like this:
```vue
<script lang="ts">
import { type SelectProps } from 'naive-ui'
type SelectThemeOverrides = NonNullable<SelectProps['themeOverrides']>

const themeOverrides: SelectThemeOverrides = {
  peers: {
    InternalSelection: {
      // single select
      paddingSingle: '0 6px 0 6px',
      // multiple select
      paddingMultiple: '0 6px 0 6px',
    }
  }
}
</script>
<template>
  <n-select
    :theme-overrides="themeOverrides"
  />
</template>
```

ksddesigner:
When will these changes be published? Pleeeease

07akioni:
Select's current behavior is by design.

PR #4487, opened 2023-02-15
feat(Code): add focus-line feature

### LINKED ISSUE #4482: n-code add line highlight
Inspired by vitepress, it provide some awesome code decoration, like [line highlight](https://vitepress.vuejs.org/guide/markdown#line-highlighting-in-code-blocks) and other features like(line focus, color diff)
### Expected API (期望的 API)
n-code can provide an array type props to specify which line should be highlight, but sometimes user don't know the specific code(code is from props or read from files), in this scene we can consider to provide a grammar which can used in code string, and then n-code process it to get the line highlight information.
Line focus and color diff are same as line highlight.
I can contribute to it.

faga295:
close #4482 
image

The `lineToDiv` mentioned in docs works by splitting the code in lines.
Highlightjs doesn't provide removePlugin api，so we can't add plugin inside component(because if we do this, each time user mount a n-code component, hljs will be added a plugin)

### DIFF
--- a/src/code/demos/enUS/focus-line.demo.vue
+++ b/src/code/demos/enUS/focus-line.demo.vue
@@ -1,0 +1,43 @@
+<markdown>
+# Focus line
+
+It can focus one line.If you want to use focus line feature, please add `lineToDiv` plugin to `hljs`.
+</markdown>
+<template>
+  <div style="overflow: auto">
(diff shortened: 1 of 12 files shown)


07akioni:
I think it's better to add background color to specific line rather than blur rest lines.

PR #4511, opened 2023-02-21
fix: FadeInExpandTransition throw mode=undefined error,which cause vue warning.

chaozwn:
[…e warning.](fix: FadeInExpandTransition throw mode=undefined error,which cause vue warning.)
below is error:
image


amadeus711 (approved):

07akioni:
fixed in other PR

PR #4522, opened 2023-02-23
Added some notes on using Notification in UMD mode

yarin-zhang:
Added some instructions on using notifications in UMD mode with related use cases. Please correct any errors.
#3772


amadeus711 (changes requested):
Pure document modification needs to `checkout` from `docs` and initiate PR merge to `docs`.

07akioni:
这应该不是必需品，只要 provider 正确的配置就可以了

PR #4727, opened 2023-04-15
chore: update deps

Sepush:
use import type for type import

update almost deps vite rollup vue ts eslint ...

pin ci node version to 18.15 for this issue https://github.com/nodejs/node/issues/47563


Sepush:
still WIP need make eslint happy

PR #4839, opened 2023-05-21
fix(data-table): "onUpdateSorter" type error

YXZT:


07akioni:
It's by design (although doesn't behavoirs correct in all scenes).

If we don't write like this, user will facing  type error if they write like this:

```
const onUpdateSorter = (sortState: SortState) => void

<n-data-table @update-sorter="onUpdateSorter" />
```

PR #4865, opened 2023-05-28
fix(time-picker): trigger an incorrect value when click the 'now' button by passing a zero value as the time step

### LINKED ISSUE #4859: time-picker's now action timestamp(seconds) error when using minutes props
2.34.4

### Vue version (Vue 版本)
3.2.36

Chrome(113.0.5672.126)

MacOS(13.3.1)

### Reappearance link (重现链接)
https://codesandbox.io/s/time-picker-error-v4vxw4?file=/src/Demo.vue

### Reappearance steps (重现步骤)
When use props minutes, Click now button will have second error
First click:
timestamp: 1685068208106
LocaleTimeString: 上午10:30:08 

Second click:
timestamp: 1685068217106
LocaleTimeString: 上午10:30:17 

### Expected results (期望的结果)
Expected two clicks have a same return timestamp, and LocaleTimeString should be 上午10:30:00

### Actual results (实际的结果)
Not reset seconds, even seconds props = 0，
If not reset seconds，after click now actions and select any time options will inaccurately
如果沒有重置秒數，則按下 now 按鈕後任何選擇時間的操作將失準

### Remarks (补充说明)
Currently I am changing update-value event to reset seconds

```ts
const handleUpdateValue = (timestamp: number) => {
  const date = new Date(timestamp);
  date.setSeconds(0, 0);
  const updatedTimestamp = date.getTime();
  emits('update-value', updatedTimestamp);
};
```

jizai1125:
closes #4859

### DIFF
--- a/CHANGELOG.en-US.md
+++ b/CHANGELOG.en-US.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## NEXT_VERSION
+
+### Fixes
+
+- Fix `n-time-picker` trigger an incorrect value when click the 'now' button by passing a zero value as the time step, closes [#4859](https://github.com/tusen-ai/naive-ui/issues/4859).
+
 ## 2.34.4
 
 ### Fixes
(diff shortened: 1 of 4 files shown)


07akioni:
`seconds` prop should not be `0`. As step, `0` is not a valid value.

PR #4880, opened 2023-05-31
fix(n-form): validate method throw exception while callback is provided

AdoKevin:
When using the `Form`, I encountered a similar issue as described in https://github.com/tusen-ai/naive-ui/issues/4823. It is possible to disable the overlay, but the root cause is still the exception thrown by the `validate `method.

Currently, the behavior of the `form.validate` method is even using a callback, the method still throws an exception. In my opinion, exception should only be thrown when using the `Promise ` way, and developers should handle exceptions themselves using `try/catch` or` .catch`. When using the callback way, there should be no need to throw exceptions. This approach is also followed by [Element ](https://element-plus.org/en-US/component/form.html#validation).

### DIFF
--- a/src/form/src/Form.tsx
+++ b/src/form/src/Form.tsx
@@ -90,38 +90,41 @@ export default defineComponent({
       validateCallback?: FormValidateCallback,
       shouldRuleBeApplied: ShouldRuleBeApplied = () => true
     ): Promise<void> {
-      await new Promise<void>((resolve, reject) => {
-        const formItemValidationPromises: Array<
-        Promise<{
-          valid: boolean
-          errors?: ValidateError[]
-        }>
-        > = []
-        for (const key of keysOf(formItems)) {
-          const formItemInstances = formItems[key]
-          for (const formItemInstance of formItemInstances) {
-            if (formItemInstance.path) {
-              formItemValidationPromises.push(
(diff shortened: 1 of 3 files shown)


src/form/src/Form.tsx
```diff
@@ -90,38 +90,41 @@ export default defineComponent({
...
         }
-        void Promise.all(formItemValidationPromises).then((results) => {
-          if (results.some((result) => !result.valid)) {
-            const errors = results
-              .filter((result) => result.errors)
-              .map((result) => result.errors)
-            if (validateCallback) {
-              validateCallback(errors as ValidateError[][])
-            }
-            reject(errors)
```
07akioni, line 118: Why not just use an else here?

07akioni: else {
  reject(errors)
}

AdoKevin: From a functional standpoint, simply adding an else statement would suffice. The reason for this modification is to prevent the program from retaining a Promise instance that may never reject or resolve, while also centralizing and improving the readability of the logic for determining whether to use a Promise approach. Therefore, I extract an internal function to decide whether to return `Promise`.

AdoKevin:
As  #5020 do same thing with this PR, I'll close this one.

PR #4889, opened 2023-06-01
#fix: updated docs stypo

iyoub:


07akioni:
fixed in fe6ec8fbbd9cd3060f54415d076b56e39ce71c67

The PR is closed due to mixed with non-relevant commits.

PR #5020, opened 2023-07-04
feat(form):  add `warningOnly` support for validation rule.

AdoKevin:
Some form item values may be considered abnormal but acceptable, for example, when an body height field has a value greater than 220cm. We want to mark this fields as warning fields. In this PR, add the `warningOnly` property to the validation rule will work.

The `warningOnly` property will display a warning feedback when the rule is unmatched, but calling the validate method will not throw an exception or resolve errors.

BTW, I'm not sure if ‘warningOnly’ is a good property name,  I am open to any other ideas and suggestions.

```JavaScript
rules: {
  count: [
    {
      trigger: ["input", "blur"],
     // show warning while submit but `validate` method will not throw exception nor resolve errors.
      warningOnly: true,
      validator(_rule: FormItemRule, value: number) {
        if (value >= 220) {
          return new Error("Really?");
        }
        return true;
      },
    },
  ];
}

```


liuzw2579:
这样 `throw?: boolean = true` ，仅提示时 `throw: false` ？

AdoKevin:
> 这样 `throw?: boolean = true` ，仅提示时 `throw: false` ？

好像也不错， `throwable` 可能更好一点点。

07akioni:
感觉还说得过去吧，那就合并了，没想到改动还不小

src/form/src/Form.tsx
```diff
@@ -108,17 +106,21 @@ export default defineComponent({
...
+            .map((result) => result.warnings)
+          if (validateCallback) {
+            validateCallback(
+              errors?.length ? (errors as ValidateError[][]) : undefined,
+              warnings?.length ? (warnings as ValidateError[][]) : undefined
+            )
           } else {
-            if (validateCallback) validateCallback()
-            resolve()
+            formInvalid ? reject(errors) : resolve()
```
07akioni, line 123: 这里有可能导致一直没有 settle

AdoKevin: 是在说传入了`validateCallback` 同时又使用await 的情况下？
这里是故意这样实现的，原因是为了解决这个issue https://github.com/tusen-ai/naive-ui/issues/4823. 
之前也为这个问题提过PR  https://github.com/tusen-ai/naive-ui/pull/4880。

我理解应该不太会有同时传入callback又使用await 情况。

PR #5025, opened 2023-07-05
fix(image):  `fallback-src` prop not working when image load error in  lazy

jizai1125:


07akioni:
seem to be duplicate of #5026

jizai1125:
> seem to be duplicate of #5026

不一样，这个是  image 组件

PR #5051, opened 2023-07-13
feat(n-tree): add `disabled-node-props-click` prop

jahnli:
在 `disabled` 的情况下还可以触发 `node-props`  `click`，提供 `disabled-node-props-click` ， 选择是否触发 `node-props`  点击事件


07akioni:
Not a proper feature. `node-props` is designed to be a DOM level props generator. It's not relevant with component's logic.

PR #5061, opened 2023-07-14
feat(date-picker): support type="week"

### LINKED ISSUE #4525: 能否在DatePicker的type选项中添加“week”选项
平常的需求中，会遇到选择周的情况。所以，请问下能否添加这个需求，非常感谢。
### Expected API (期望的 API)
<n-date-picker type="week" clearable />

suica:
fixes #4525 

## Todos
- [x] finish `type="week"`
- [ ] check disabled style
- [x] check the i18n files for adding new localized text
- [ ] check the enUS doc
- [x] update the changelog
- [ ] clean the code


src/date-picker/src/DatePicker.tsx
```diff
@@ -387,6 +391,10 @@ export default defineComponent({
           result.push('confirm')
           return result
         }
+        case 'week': {
+          result.push('confirm')
```
suica, line 395: Maybe we also don't need this

Sepush:
demos for both en-US and zh-CN but you can add it finally

Sepush:
when impl range need disable the date before the start time

Sepush:
in week mode we do not need the hover style for the speific day 
image
the value/the arg for `on-update:value` the day of firt day of week not the click 
also need handle the `default-time` prop

suica:
Since type `date` and `week` share a simlilar outlook, most code could be resused. So i chose to extend `Date.tsx` to make it support an extra `week` mode.
Do you think it is a good idea? Should we copy the code and make a new week panel?

src/date-picker/src/DatePicker.tsx
suica: Removed. Now, `week` and `date` share the same case branch.

suica:
> demos for both en-US and zh-CN but you can add it finally

Added.



> in week mode we do not need the hover style for the speific day image the value/the arg for `on-update:value` the day of firt day of week not the clicked day also need handle the `default-time` prop

Fixed. Removed hover effect for both `--slightly-covered` and `--heavily-covered` modifiers.

@Sepush And, do you think we need a new dedicated file for `week` panel? Now extending the original date panel seems to be a little bit awkward...

Sepush:
> Since type `date` and `week` share a simlilar outlook, most code could be resused. So i chose to extend `Date.tsx` to make it support an extra `week` mode. Do you think it is a good idea? Should we copy the code and make a new week panel?

Just do it if need to

AdamsTsui:
Good job，Thanks!

FlareZh:
Any progress?

Sepush:
cc @07akioni  need code review

suica:
Hi, @07akioni ! i need your general comment on this implementation to make sure I am on the right track, so I can finish the rest todos :)

Sepush:
Thank you for your contribution. @suica 
would you mind add a week number row in the panel with gray color and a dot `.` like  `26.`
I think we can do it in another PR.

deiiviidas:
also need week days translate for example
near weekFormat, [weekDaysFormat ??]

weekDayLong =>[ 'Monday' , ....]
weekDayShort =>[ 'Mon' , ...]

PR #5062, opened 2023-07-14
fix(clear): invoke clear on mousedown

### LINKED ISSUE #4658: Auto-Complete behavior
2.34.3

### Vue version (Vue 版本)
3.3.0-alpha.5

Chrome (111.0.5563.147)

Windows10

### Node version (Node 版本)
16.19.1

### Reappearance link (重现链接)
https://codesandbox.io/s/agitated-sea-79166k

### Reappearance steps (重现步骤)
1. Input something
2. Click outside
3. Click clear icon to clear the content
4. The select menu show up for a tick and disappear

### Expected results (期望的结果)
After click the clear icon, select menu should not show up anymore

### Actual results (实际的结果)
The select menu flashed for a tick and disappear, I think it's bad user experience

### Remarks (补充说明)
I think it's not a design-so behavior right?
I've looked deeper and find out that it's the n-input default behavior that the input would auto-focus after clicked clear icon (in the n-input `handleMouseDown` function).
Do you guys think it's ok to add a flag for n-input like `focusAfterClear` which is default true, and in auto-complete, just pass `:focus-after-clear="false"` to n-input and the behavior is gone.
I can make a pr, with tested case and document added. Just wondering how you guys think about it.

suica:
fixes #4658


Sepush:
pick a demo add `clearable` and changlog for both `CHANGELOG.en-US` and `CHANGELOG.zh-CN`

PR #5063, opened 2023-07-14
feat(tabs): add animation for segment

### LINKED ISSUE #3910: Add tab animation for `<NTab>` of `segment` type
### The problem this function solves (这个功能解决的问题)

When using `<NTab>` of `line` type, there is a line movement animation. which is nice, but not for the `segment` type, it looks a little bit stiff when switching the active tab.

### Expected API (期望的 API)

No extra API needed.

suica:
closes #3910

Just a ugly brute force implementation, for prototyping an idea. The following things need to be checked:
- [ ] resize
- [ ] style

Demo:

tab

### DIFF
--- a/src/tabs/src/Tabs.tsx
+++ b/src/tabs/src/Tabs.tsx
@@ -516,18 +516,30 @@ export default defineComponent({
     })
 
     const tabsRailElRef = ref<HTMLElement | null>(null)
-    watch(mergedValueRef, () => {
-      if (props.type === 'segment') {
-        const tabsRailEl = tabsRailElRef.value
-        if (tabsRailEl) {
-          void nextTick(() => {
-            tabsRailEl.classList.add('transition-disabled')
-            void tabsRailEl.offsetWidth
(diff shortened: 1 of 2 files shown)


src/tabs/src/Tabs.tsx
```diff
@@ -389,6 +389,33 @@ export default defineComponent({
...
+    const tabsRailElRef = ref<HTMLElement | null>(null)
+    const segmentCapsuleElRef = ref<HTMLElement | null>(null)
+
+    function updateSegmentPosition (): void {
+      if (props.type === 'segment') {
+        const { value: tabsRailEl } = tabsRailElRef
+        if (!tabsRailEl) return
+        void nextTick(() => {
+          tabsRailEl.classList.add('transition-disabled')
+          const activeTabEl = tabsRailEl.querySelector('.n-tabs-tab--active')
```
07akioni, line 401: The way that select active tab need to be refactored, query selector is not preferred.

07akioni:
check refactor in b843bea49

PR #5124, opened 2023-08-02
fix(Input): Input props onKeydown is Array

liulinboyi:
## Issue
If use like this:
```vue
<template>
  <n-input
    @keydown.ctrl.c="handleDownKey"
    @keydown.ctrl.a="handleDownKey"
  />
</template>

<script setup lang="ts">
import { NInput } from "naive-ui";

const handleDownKey = () => {
  console.log("key down pressed");
};
</script>
```
will throw Error
image

the reason is the `props.onKeydown` maybe `Array`.

## Link
https://codesandbox.io/p/sandbox/blissful-rubin-wz7sfn?file=%2Fsrc%2FApp.vue%3A16%2C1

### DIFF
--- a/src/input/src/Input.tsx
+++ b/src/input/src/Input.tsx
@@ -659,7 +659,7 @@ export default defineComponent({
       on('mouseup', document, hidePassword)
     }
     function handleWrapperKeydown (e: KeyboardEvent): void {
-      props.onKeydown?.(e)
+      if (props.onKeydown) call(props.onKeydown, e)
       switch (e.key) {
         case 'Escape':
           handleWrapperKeydownEsc()


src/input/src/Input.tsx
```diff
@@ -659,7 +659,18 @@ export default defineComponent({
       on('mouseup', document, hidePassword)
     }
     function handleWrapperKeydown (e: KeyboardEvent): void {
-      props.onKeydown?.(e)
+      if (props.onKeydown) {
```
amadeus711, line 662:
```suggestion
      if (props.onKeydown) call(props.onKeydown, e)
```

amadeus711 (changes requested):

src/input/src/Input.tsx
liulinboyi: ok，Cool！

amadeus711 (approved):

PR #5127, opened 2023-08-03
fix(input): autofill's default background color

### LINKED ISSUE #5123: An API to disable autofill background color for n-input components

This feature request is mainly targeted to the behavior after the _**autofill**_ feature of the browser is activated.

When a n-input component is autofilled, browsers will add a special background color to the input box through User Agent CSS. 

Problem is that the added background color does not suit some of the n-input component very well, for example when the input box has a prefix or suffix icon, or when the input box is in dark mode. They look a little bit weird.

Something like this: I believe the autofilled white background color should expand to the whole input box, but here the background of the suffix icon is still dark.
image

I am looking forward to an API that enables me to modify the autofill background color, or just simply disable it.

Also if naive-ui would like to work on this problem and make the autofill background expand to cover the whole input box, that would be nice too!

### Expected API (期望的 API)

```
<n-input placeholder="pswd" autofillbg="disabled">
```

jizai1125:
closes #5123


src/input/src/styles/input.cssr.ts
```diff
@@ -92,7 +92,11 @@ export default cB('input', `
     `),
     c('&:-webkit-autofill ~', [
       cE('placeholder', 'display: none;')
-    ])
+    ]),
+    // overlay autofill's default background color
+    c('&:autofill, &:-webkit-autofill, &:-webkit-autofill:hover, &:-webkit-autofill:focus,&:-webkit-autofill:active', `
```
07akioni, line 97:
```suggestion
    c('&:autofill, &:-webkit-autofill, &:-webkit-autofill:hover, &:-webkit-autofill:focus, &:-webkit-autofill:active', `
```

PR #5248, opened 2023-09-18
Space reverse

kvii:
Add support for `flex-direction: row-reverse` in NSpace.

## use case

User can use `<NSpace vertical reverse>` with action buttons, so that users will have a correctly focus order.

image


kvii:
No test for this feature. Because I don't know how to write testing code for ui component.
And my code style also not good.
You can implement it by your self, then close my PR.

PR #5438, opened 2023-12-06
feat(input): expose `clear` method

### LINKED ISSUE #5423: 表单组件暴露clear方法
建议如input、select等带有clearable属性的表单组件暴露clear实例方法
### Expected API (期望的 API)
InputInst.clear()

jizai1125:
closes #5423


src/input/src/Input.tsx
```diff
@@ -812,6 +812,7 @@ export default defineComponent({
       inputElRef,
       textareaElRef,
       isCompositing: isComposingRef,
+      clear: handleClear,
```
07akioni, line 815: To conform exposed type. It should be `clear: () => handleClear()`. Also, `onClear` shouldn't be called since it's an user reaction.

PR #5476, opened 2023-12-19
fix(drawer): handleMaskClick event runs the risk of being executed multiple times

eivmosn:

### DIFF
--- a/src/drawer/src/Drawer.tsx
+++ b/src/drawer/src/Drawer.tsx
@@ -224,6 +224,18 @@ export default defineComponent({
       if (onMaskClick) onMaskClick(e)
     }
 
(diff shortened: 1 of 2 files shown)


07akioni:
I've added new implementation, the current one may not be proper.

eivmosn:
> I've review the current implementation & seems it won't be executed twice?

Try the demo multiple drawers. If the first drawer component set property 'show-mask' to false, and then open drawer which is nested in, and it will close both of it.

eivmosn:
> I've review the current implementation & seems it won't be executed twice?

please check https://stackblitz.com/edit/vitejs-vite-ts5wmu?file=src%2FApp.vue

PR #5521, opened 2024-01-01
docs: fix checkbox indeterminate demo can not be displayed correctly

XiaoDaiGua-Ray:


07akioni:
> I think the example is really a bit opaque. I think we should replace the checked and intermidiate buttons with controlled switches. That way you can also see how clicking changes different states.

You are correct. I've update the example in 40c2068b

PR #5548, opened 2024-01-08
fix(form): `validate` unable to complete execution properly

tiny-craft:


tiny-craft:
Sorry that I make mistake for the comment info. I will re-commit a pr later.

PR #5549, opened 2024-01-08
fix(form): method 'validate' unable to complete execution properly

tiny-craft:


tiny-craft:
It seems that this bug already fixed in v2.37.1, great~

Issue #5575, opened 2024-01-10
Hydration mismatch issues with SSR


becem-gharbi:
### TuSimple/naive-ui version (版本)
2.37.3

### Vue version (Vue 版本)
3.4.6

### Browser and its version (浏览器及其版本)
Chrome(120.0.6099.216)

### System and its version (系统及其版本)
Linux(5.0)

### Node version (Node 版本)
18.18.0

### Reappearance link (重现链接)
https://stackblitz.com/edit/github-euic2u?file=nuxt.config.ts

### Reappearance steps (重现步骤)
Run `npm run dev` and check console.

### Expected results (期望的结果)
No hydration issues.

### Actual results (实际的结果)
Multiple <[Vue warn]: Hydration attribute mismatch> warnings are shown.

### Remarks (补充说明)
This issue is related to vue v3.4.0 introducing attributes check for hydration mismatch [PR](https://github.com/vuejs/core/pull/5953).

I have already opened an [issue](https://github.com/vuejs/core/issues/10057) on the vue/core repo. ~However these type of issues are taken as special cases. Thus it's better to fix the library's implementation instead.~

I have tested almost all the components. The issues are related to the following ones
- `n-input` attribute `readonly` 
- `n-checkbox` attribute `id` of the label
- `n-image` attribute `height` and `width`
- `n-tree` attribute `draggable`

2556227850:
问题解决了麻烦艾特我一下

becem-gharbi:
These issues are getting solved in the vue core that's why I'm closing this one.

fogine:
This is still an issue for case of `n-checkbox` `id` attribute of the label. `naiveui version: ^2.37.3`

When rendering `n-checkbox` on server with a label it generates random `id` and assigns it to `n-checkbox` `aria-labelledby=` and `id=` arguments of `n-checkbox__label` span respectivelly.
I tried to assign static value to `aria-labelledby=` on the `n-checkbox` element which does nothing as its not supported.
It should be possible to get around this issue.

![img-2024-02-20-122100](https://github.com/tusen-ai/naive-ui/assets/7884288/5e3e5f58-55a1-4cc5-9116-c7e4cdcb1c10)

becem-gharbi:
Probably the hydration mismatch of the `id` attribute is going to be solved in `vue` v3.5. A great video from a nuxt team member points to this issue https://youtu.be/d4-OjlicbiY?t=367.

fogine:
[...]

PR #5578, opened 2024-01-10
fix(input): Fix readonly mismatch on SSR

becem-gharbi:
solves(#5575)

If the attribute `readonly` is present it will be checked for hydration mismatch on `vue` >=v3.4.0  [source](https://github.com/vuejs/core/blob/a3725a729cc99dc0ebe7c50ca65f183b8ff30073/packages/runtime-core/src/hydration.ts#L761
).

The issue is related to the fact that the `key` represents the attribute name which is `readonly`. However the key of the HTML element is `readOnly`. This results to an incorrect evaluation when value is `false`. 

The proposed solution is to ignore the `readonly` attribute when its value is `false`.

### DIFF
--- a/src/input/src/Input.tsx
+++ b/src/input/src/Input.tsx
@@ -1126,7 +1126,12 @@ export default defineComponent({
                         disabled={this.mergedDisabled}
                         maxlength={countGraphemes ? undefined : this.maxlength}
                         minlength={countGraphemes ? undefined : this.minlength}
-                        readonly={this.readonly as any}
+                        {...Object.assign(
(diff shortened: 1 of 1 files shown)


OrbisK:
isnt this fixed with https://github.com/vuejs/core/commit/972facee0d892a1b6d9d4ad1da5da9306ed45c3f?

becem-gharbi:
This is now solved on vue v3.4.8 https://github.com/vuejs/core/blob/main/CHANGELOG.md#348-2024-01-10.

Issue #5623, opened 2024-01-23
面板分割 Split组件的resize-trigger的样式显示问题


kirklin:
### Describe the bug

面板分割 Split组件的resize-trigger的样式显示不出来

### Steps to reproduce

仓库地址：
https://github.com/kirklin/celeris-web/blob/master/apps/admin/src/pages/chat/components/ChatPanel/index.vue
部署后地址：
https://celeris-web.vercel.app/#/chat/index

### Link to minimal reproduction

https://github.com/kirklin/celeris-web/blob/master/apps/admin/src/pages/chat/components/ChatPanel/index.vue

### System Info

```Shell
System:
    OS: macOS 13.5
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 578.73 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.2.0 - ~/.nvm/versions/node/v20.2.0/bin/node
    npm: 9.6.6 - ~/.nvm/versions/node/v20.2.0/bin/npm
    pnpm: 8.10.5 - ~/.nvm/versions/node/v20.2.0/bin/pnpm
    bun: 1.0.3 - ~/.nvm/versions/node/v20.2.0/bin/bun
  Browsers:
    Chrome: 120.0.6099.234
    Safari: 16.6
```

### Used Package Manager

pnpm

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/tusen-ai/naive-ui/blob/main/CONTRIBUTING.zh-CN.md).
- [X] Read the [docs](https://www.naiveui.com/zh-CN/).
- [X] Check that there isn't [already an issue](https://github.com/tusen-ai/naive-ui/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/tusen-ai/naive-ui/discussions).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

XueHua-s:
In the code you provided, I don't see the attribute of resize-trigger in the member of v-slots.

kirklin:
I would like to use the default rather than the custom

kirklin:
> I would like to use the default rather than the custom

[...]

PR #5670, opened 2024-02-06
fix(split): missing losing css variables with inline theme disabled (…

kirklin:
closes(#5623)


07akioni:
I've fixed it. Will release it in next version.

kirklin:
I updated to 2.38.0 and this problem seems to persist. I'm applying for this PR to be reinstated

jahnli:
指的应该是 `2.38.0` 后的一个版本

PR #5718, opened 2024-02-25
fix(split): missing losing css variables with inline theme disabled

kirklin:
closes(#5717)


07akioni:
already fixed in latest version

PR #5773, opened 2024-03-12
docs(popover): `title` prop removed

jizai1125:


src/upload/src/interface.ts
```diff
@@ -124,7 +124,7 @@ export interface XhrHandlers {
 
 export interface UploadInst {
   openOpenFileDialog: () => void
-  submit: () => void
+  submit: (fileId?: string) => void
```
woodhull, line 127: How is this change related?

jizai1125: I changed the title and consolidated all the fixes into one PR to avoid making another.

PR #5974, opened 2024-05-23
✨ feat(select): Option to keep the label of the selected item if it i…

keypadDancer:
Retaining a label makes it easy to make changes to the original label to quickly re-filter the filter, or to copy the text from the label and use it in other ways

This feature really makes a lot of sense believe me
Hopefully it will be merged soon

If that doesn't work, I hope it's possible to return the patternRef variable from the select


keypadDancer:
Video_2024-05-23_143145

keypadDancer:
https://github.com/tusen-ai/naive-ui/issues/5407

keypadDancer:
It seems to have been written more than once.
beingCreatedOptionsRef.value = emptyArray

07akioni:
> 大佬，把这个pr关掉吧，我发现我想要的也不是这样的，和这个会有一点差别，我会重新提交一个新的pr Let's turn this pr off, I realized that's not what I want either, it will be a little different from this one, I'll resubmit a new pr

OK, I just now saw this PR. If you find out what you need feel free to open a new one.

PR #6007, opened 2024-06-02
feat(`n-date-picker`): add event support for shortcuts

jahnli:
close https://github.com/tusen-ai/naive-ui/issues/6001


07akioni:
This is not an expected implementation.

The original issue is ambiguous. I don't think this is the proper solution.

07akioni:
We will find a new solution after that issue is made clear.

Issue #6036, opened 2024-06-09
n-upload GET request with headers


swuecho:
### Clear and concise description of the problem

![image](https://github.com/tusen-ai/naive-ui/assets/666683/b9e97302-145b-40cd-abd5-bb8b1c0084ae)

点击这个link 触发下载, 但是headers 没有带上, 导致 unauthorized

### Suggested solution

GET request 带上headers

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/tusen-ai/naive-ui/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://www.naiveui.com/en-US).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

swuecho:
或者提供是否显示文件链接的props, 显示了文件链接, 但是又不能用, 比较棘手.

jahnli:
没太理解需求    `data`    `with-credentials`  ?

swuecho:
点击链接的时候会发送get 请求. 后台根据JWT token 鉴权的, with-credentials 根据文档是带上cookies

swuecho:
https://github.com/tusen-ai/naive-ui/blob/6bea68a3a55a832dc6f90d79507cdd096ea8d0e6/src/upload/src/UploadFile.tsx#L261

https://github.com/tusen-ai/naive-ui/blob/6bea68a3a55a832dc6f90d79507cdd096ea8d0e6/src/upload/src/UploadFile.tsx#L179

handlePreviewClick 应该 preventDefault() ? 否则 <a href> 默认的行为会打开链接?

swuecho:
> 点击链接的时候会发送get 请求. 后台根据JWT token 鉴权的, with-credentials 根据文档是带上cookies

这个猜想有误, 应该是浏览器打开了链接.

jahnli:
可能设计的点击浏览器打开

PR #6049, opened 2024-06-13
fix: upload on preview prevent default

swuecho:
fix: https://github.com/tusen-ai/naive-ui/issues/6036


07akioni:
We'd better just pass this event object to `onPreview`, since user may just want to log something, not take over the preview process.

PR #6057, opened 2024-06-14
docs(vitepress): fully `ts` and fixed a problem with dark mode syncing

majoson-chen:
- 对 ts 添加完全支持
- 现在可以与 vitepress 的暗黑模式同步了


07akioni:
1. I think keep `mts` extension is better since it's supported in both module type.
2. If ts is needed, I think it should be a parallel section for js section, not simply replace it.
3. Code block indent is not correct.

PR #6060, opened 2024-06-17
fix(tabs): vertical mode addable exception

liuyang0826:
修复tabs组件垂直模式添加按钮及前后缀表现异常https://github.com/tusen-ai/naive-ui/issues/6059
image

### DIFF
--- a/src/tabs/demos/zhCN/placement.demo.vue
+++ b/src/tabs/demos/zhCN/placement.demo.vue
@@ -19,6 +19,7 @@
       :key="type + placement"
       :type="type"
       animated
+      addable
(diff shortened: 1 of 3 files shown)


src/tabs/demos/zhCN/placement.demo.vue
```diff
@@ -19,6 +19,7 @@
       :key="type + placement"
       :type="type"
       animated
+      addable
```
07akioni, line 22:
```suggestion
```

07akioni:
The issue fixes some problems but not all of them, so I create a commit that fixes them.

PR #6094, opened 2024-06-28
fix(n-input-number): exception when the value is a string in precisio…

jahnli:
close https://github.com/tusen-ai/naive-ui/issues/6091

### DIFF
--- a/src/input-number/src/utils.ts
+++ b/src/input-number/src/utils.ts
@@ -36,8 +36,6 @@ export function format (
 ): string {
   if (typeof value !== 'number') return ''
   return precision === undefined ? String(value) : value.toFixed(precision)
-    ? String(value)
-    : Number(value).toFixed(precision)
 }
 
 export function parseNumber (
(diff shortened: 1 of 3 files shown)


07akioni:
I think the proper way to solve the problem is to view non-number value as undefined or null. Force casting value into `number` may causing a NAN problem.

src/input-number/src/utils.ts
```diff
@@ -35,7 +35,9 @@ export function format (
   precision: number | undefined
 ): string {
   if (value === undefined || value === null) return ''
-  return precision === undefined ? String(value) : value.toFixed(precision)
+  return precision === undefined
```
07akioni, line 38:
```suggestion
  return precision === undefined ? String(value) : value.toFixed(precision)
```

src/input-number/src/utils.ts
```diff
@@ -34,8 +34,10 @@ export function format (
   value: number | undefined | null,
   precision: number | undefined
 ): string {
-  if (value === undefined || value === null) return ''
+  if (typeof value !== 'number') return ''
   return precision === undefined ? String(value) : value.toFixed(precision)
+    ? String(value)
+    : Number(value).toFixed(precision)
```
07akioni, line 40:
```suggestion
```

PR #6132, opened 2024-07-10
feat(locales): add Uzbek language support

azabroflovski:
## Description 

Added support for Uzbek locale

p.s with love from Tashkent ❤️ (Uzbekistan). We are eager to further support this locale in case of updates.

Thank you, Naive UI team, for your hard work and contributions. Our team believes that this is the best UI library for Vue and we are successfully using it in production.


07akioni:
Thanks for your contribution. Your comment inspired me a lot.

PR #6225, opened 2024-08-02
fix: use correct pattern for ISO weekFormat

JianJroh:
Should be use `RRRRR` ( ISO week-numbering year ) and `I` (ISO week of year) as default `weekFormat`. See [date-fns docs](https://date-fns.org/v3.6.0/docs/format ).

Another issue, if we use ISO week date we should diable `first-day-of-week` prop for week picker because ISO-week start with Monday. See [wiki](https://en.wikipedia.org/wiki/ISO_week_date#:~:text=Weeks%20start%20with%20Monday%20and%20end%20on%20Sunday.) , [HTML week concept](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#concept-week).
 Different  `first-day-of-week` can break the consistency  of week selection on panel with the week-formatted text.

### DIFF
--- a/src/locales/common/arDZ.ts
+++ b/src/locales/common/arDZ.ts
@@ -31,7 +31,7 @@ const arDZ: NLocale = {
     dateFormat: 'yyyy-MM-dd',
     dateTimeFormat: 'yyyy-MM-dd HH:mm:ss',
     quarterFormat: 'yyyy-qqq',
-    weekFormat: 'RRRR-w',
+    weekFormat: 'RRRR-I',
     clear: 'مسح',
     now: 'الآن',
     confirm: 'تأكيد',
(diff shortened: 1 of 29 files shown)


07akioni:
I think by default we should use local week for date picker, since most people would use their common sense to determine what is current week.

CleanShot 2024-08-06 at 00 03 59@2x

If there are inconsistent issues I think we should solve them first without using ISO formatted week.

JianJroh:
> I think by default we should use local week for date picker, since most people would use their common sense to determine what is current week.
> 
> CleanShot 2024-08-06 at 00 03 59@2x
> 
> If there are inconsistent issues I think we should solve them first without using ISO formatted week.

I try to resolve it in new [pr](https://github.com/tusen-ai/naive-ui/pull/6234)

PR #6235, opened 2024-08-06
feat(time): add support for custom date-fns format options

kuoruan:
添加了一个 `formatOptions` 选项，支持自定义 date-fns 格式化参数。

See: https://date-fns.org/v2.29.3/docs/format#arguments


src/time/src/Time.ts
```diff
@@ -27,7 +28,11 @@ export const timeProps = {
   unix: Boolean,
   format: String,
   text: Boolean,
-  timeZone: String
+  timeZone: String,
+  formatOptions: {
+    type: Object as PropType<FormatOptions>,
+    default: undefined
+  }
```
07akioni, line 35:
```suggestion
  formatOptions: Object as PropType<FormatOptions>
```

src/time/src/Time.ts
```diff
@@ -79,7 +93,7 @@ export default defineComponent({
         return mergedFormatRef.value(
           mergedTimeRef.value,
           props.format,
-          dateFnsOptionsRef.value
+          formatOptionsRef.value
```
07akioni, line 96: line 103 & line 110 miss changes.

Also, if `dateFnsOptionsRef ` is only referred once after changes applied. We should remove this var and change it to inline code.

kuoruan: I used to thought the `formatOptions` should only work when the `format` props is satisfied.
You are right, we should add it to line 103 and 110 to align the options.

kuoruan:
@07akioni Changed, need review.

PR #6236, opened 2024-08-06
feat(date-picker): add `close-month-panel-on-select` prop to close the month panel after the user has selected the month

JianJroh:
After setting `close-month-panel-on-select`:

20240806160838_rec_


07akioni:
We need a prop to control this behavior, not simply adjust it.

07akioni:
If we just change the behavior to hide month panel after clicking month item. Some people would say they want to select year after month selection.

kuoruan:
Agreed, add an option to control it is better.

PR #6375, opened 2024-09-23
fix: (n-dialog) props types fixes and new async example

Hackenator3000:


OrbisK:
Hey, thank you for your PR. 

Can you briefly explain why this is necessary? Can't you just pass the ref value? Then everything is reactive, isn't it?

OrbisK:
I think you dont need to change the types of the props. You should change the types of the composable options

Hackenator3000:
> You should change the types of the composable options
Which of composables?

> I think you dont need to change the types of the props
useDialog already works with reactive props I just extended some props to fix type errors and its don't break anything

OrbisK:
> > You should change the types of the composable options
> 
> Which of composables?
> 
> > I think you dont need to change the types of the props
> 
> useDialog already works with reactive props I just extended some props to fix type errors and its don't break anything

Mhh. I understand what you mean. But that would mean that we would have to wrap every prop in `MaybeRefOrGetter`. I still think that this should be done via the composable. But in the end, that's not my decision either.

@07akioni what do you mean?

PR #6421, opened 2024-10-01
feat(data-table): add animations for expand icon

jahnli:


07akioni:
Since expansion doesn't has animation, we don't add animation for expand icon.

PR #6790, opened 2025-03-03
Clarify Docs on Applying Dark Theme Globally

abhishek-junghare:
Setting the dark theme globally can be confusing, or at least, it was for me. When enabling dark mode, users typically expect it to apply globally, but this isn’t the case by default. I spent over an hour troubleshooting before finally discovering the relevant information in the [Sync style of the body element](https://www.naiveui.com/en-US/dark/docs/customize-theme#Sync-style-of-the-body-element) section. However, this section is at the bottom of the documentation and isn't titled clearly, making it easy to miss.

I propose two possible solutions:

1. Rewrite and reposition the [Sync style of the body element](https://www.naiveui.com/en-US/dark/docs/customize-theme#Sync-style-of-the-body-element) section. The new title could be something more direct, like "Apply Theme Globally", and the section should clearly explain that adding this step ensures global theme application. Moving it immediately after the [Use dark theme](https://www.naiveui.com/en-US/dark/docs/customize-theme#Use-dark-theme) section would make it more visible to users configuring dark mode.
2. Add a brief note about `<n-global-style>` right after the [Use dark theme](https://www.naiveui.com/en-US/dark/docs/customize-theme#Use-dark-theme) section to ensure users immediately understand what’s required.

This PR implements the second solution. If a more detailed rewrite is preferred, that might be a better long-term fix. This small change should help prevent confusion and improve the developer experience.


jahnli:
image

There seems to be a problem, containing redundant commits

PR #6883, opened 2025-05-21
chore: change button to setup api

7hao123:
this is a test file


jahnli:
Because this is a test file, I will turn it into a draft first.

PR #6947, opened 2025-06-18
fix(n-menu): disabled style not working

blackawn:
修复 `n-menu` 在父节点设置 `disabled`，子节点为 `type: "group"` 的禁用样式失效，关闭 [#6792](https://github.com/tusen-ai/naive-ui/issues/6792)


07akioni:
The original fix is not correct. Since by design,  an option can have most only one of `NSubmenu` and `NMenuOptionGroup` injection.

I've updated the implementation in 6df803d21ba240dfd556a4a9b8e407587d36112c

PR #7093, opened 2025-07-17
feat(date-picker): add custom default time function support

NovaStream2030:
Enhanced the `default-time` prop in DatePicker components to support custom functions for dynamic time calculation, in addition to the existing string and array formats. Fully backward compatible - existing string and array formats continue to work as before.

**Use Cases**

This enhancement enables more flexible time handling scenarios:
- Set current time for today, specific times for other dates
- Apply different default times based on date context
- Implement business logic for time selection (e.g., working hours)


NovaStream2030:
bump?

PR #7184, opened 2025-09-03
date-picker adds 'prefix' slot

### LINKED ISSUE #6787: Add #prefix slot to date picker
### Clear and concise description of the problem

It would be great if i could pass a prefix to date picker. Prefixes and suffixes on forms allow me to avoid using topside labels and thereby make my forms look a lot cleaner. It works great with normal inputs but sadly not with date pickers

P.S. Thank you for the library, its awesome, i would not know what to do without it

### Suggested solution

Add a #prefix slot to date picker

### Alternative

_No response_

### Additional context

_No response_

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/tusen-ai/naive-ui/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://www.naiveui.com/en-US).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

mqnu00:
close #6787

### DIFF
--- a/CHANGELOG.en-US.md
+++ b/CHANGELOG.en-US.md
@@ -33,7 +33,7 @@
 - Add `n-heatmap` component.
 - - `n-image` adds showPreview methods, closes [#6695](https://github.com/tusen-ai/naive-ui/issues/6695).
 - `n-image-preview` `n-image-group` support being used independently.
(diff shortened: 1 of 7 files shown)


Dylan0916:
i need this

PR #7278, opened 2025-10-17
Pr typo

kvii:
这里忘记说”事件“了。
英文原文：The native disabled button won't trigger some mouse events so `n-popover` can't listen to it.


kvii:
搞错分支了。我重新提一个吧

PR #7294, opened 2025-10-30
fix(): date picker week labels

jakubhruby:
image


jakubhruby:
@07akioni ready for review

PR #7295, opened 2025-10-30
feat(): add some html props

jakubhruby:
For accessible web we need to be able to add some ARIA attrs directly to `<th>` cells. It can be done in the same way as cellProps for regular `<td>`s . 
Adding component prop `caption` to generate `<caption>` for the same reason. 
Adding Select loading element html props.
Adding Radio input props.
Adding Date picker input props.


jakubhruby:
@07akioni ready for review

PR #7410, opened 2026-01-07
feat(markdown): add NMarkdown

rwv:
### What
- Add NMarkdown runtime Markdown renderer (token -> VNode).
- Support code blocks with injected highlight.js (same pattern as NCode).
- Add sanitize options: `sanitize` (default true) strips unsafe URLs and escapes raw HTML; `sanitizeHtml` hook enables rendering sanitized HTML when needed.
- Add demos (basic/hljs/security) and unit tests.


jahnli:
:+1:

src/markdown/src/Markdown.tsx
```diff
@@ -0,0 +1,468 @@
+import type { Token, Tokens } from 'marked'
```
Sepush: - prefer https://unifiedjs.com/ [remarkjs](https://github.com/remarkjs) stack
- due to package size, Naive UI doesn't include the package
- the approach should be like `n-code` `n-hilight` `n-equation `

PR #7461, opened 2026-02-03
fix(`n-modal`): nested component jitter when opening modal

jahnli:
close https://github.com/tusen-ai/naive-ui/issues/4862
close https://github.com/tusen-ai/naive-ui/issues/6355

### DIFF
--- a/CHANGELOG.en-US.md
+++ b/CHANGELOG.en-US.md
@@ -1,5 +1,13 @@
 # CHANGELOG
 
+## NEXT_VERSION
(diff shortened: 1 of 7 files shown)


07akioni:
Decided not to fix this for now - browser updates might resolve this issue. I've encountered similar problems before, and the only solution was adding translate-z, but that comes with some side effects. If this is important, please provide more details, or feel free to comment here. I'll look into it further once more people need this addressed.

PR #7532, opened 2026-03-17
docs: optimize js output

micaiguai:
Optimize js output.
1. ensure import not removed.
2. keep the format.

### 1. ensure import not removed.
Before
image
After
image

### 2. keep the format.
Before
image
After
image


07akioni:
Seems the problem is fixed by using oxc-transform in the lastest state of the repo.

Issue #8079, opened 2026-05-12
Input v-model is not updated with the last composing character (Korean IME)


kyumoon:
### Describe the bug

  When typing Korean with an IME, the last in-progress character is not
  reflected on `v-model:value` until composition is explicitly ended (blur,
  Enter, switching language, etc.).

  Typing `안녕` and pausing leaves the bound state at `안` because `녕` is
  still in composition.

  ### Reproduction

  ```vue
  <script setup>
  import { ref } from 'vue'
  import { NInput } from 'naive-ui'
  const text = ref('')
  </script>

  <template>
    <n-input v-model:value="text" />
    <p>model: {{ text }}</p>
  </template>
  ```

  1. Switch input source to Korean.
  2. Type `안녕` and stop typing.
  3. The `<p>` shows `안`, not `안녕`.

  ### Expected

  The bound model should reflect the visible text in the input as soon as the
  user stops typing.

  ### Actual

  `handleInput` returns early while `isComposingRef.value === true`, so
  `doUpdateValue` is never called during composition.

  ### Root cause

  `src/input/src/Input.tsx`, inside `handleInput`:

  ```ts
  syncSource = targetValue
  if (isComposingRef.value)
    return // model update is skipped for the entire composition window
  // ...
  doUpdateValue(targetValue, { source: index })
  ```

  Only `compositionend` triggers a model update. Any character still in IME
  composition when the user stops typing is therefore missing from the bound
  state.

<img width="276" height="178" alt="Image" src="https://github.com/user-attachments/assets/3534735c-377f-4854-8771-4239f6eb3f5d" />

### Steps to reproduce

  **1. Enable a Korean IME**
  - **macOS:** System Settings → Keyboard → Input Sources → Add "2-Set Korean". Toggle with `Ctrl+Space`.
  - **Windows:** Settings → Language → Add Korean. Toggle with `Win+Space`.

  **2. Mount an `<n-input v-model:value="text" />` bound to a ref, and render `{{ text }}` next to it.**

  **3. Type `안녕`**

  With the Korean IME active, press these QWERTY keys: **`d` `k` `s` `u` `d`**

  **4. Observe**
  - Input field shows `안녕` ✅
  - Rendered model shows `안` ❌ (last character missing)

### Link to minimal reproduction

no reproduction

### System Info

```Shell
System:
    OS: macOS 15.6.1
    CPU: (12) arm64 Apple M2 Pro
    Memory: 133.45 MB / 32.00 GB
[...]

PR #8085, opened 2026-05-15
fix(input): update model value during IME composition

ruguoba:
## Problem

When typing with IME (Korean, Chinese, Japanese), the model value () is not updated during composition. This causes the model to fall out of sync with the visible input text.

**Reported in**: #8079

## Root Cause

In , the  function returns early when  is true (line 513-514):

```ts
syncSource = targetValue
if (isComposingRef.value)
  return  // model update is skipped for the entire composition window
```

This means during IME composition,  is never called, so the model value stays stale until  fires.

## Fix

Update the model value during composition while still skipping validation (), cursor control, and  to avoid interfering with IME behavior:

```ts
syncSource = targetValue
if (isComposingRef.value) {
  // Update model during composition to keep it in sync with visible input
  if (!props.pair) {
    if (event === 'input') {
      doUpdateValue(targetValue, { source: index })
    } else {
      doChange(targetValue, { source: index })
    }
  } else {
    // ... pair handling
  }
  return  // Skip validation, cursor control, and forceUpdate
}
```

## Testing

- Tested with Korean IME (2-Set Korean)
- Model value now reflects visible input during composition
- No interference with IME behavior (cursor position, composition window)

### DIFF
--- a/src/input/src/Input.tsx
+++ b/src/input/src/Input.tsx
@@ -510,8 +510,35 @@ export default defineComponent({
         }
       }
       syncSource = targetValue
-      if (isComposingRef.value)
+      if (isComposingRef.value) {
+        // Update model during composition to keep it in sync with visible input
+        // Skip validation, cursor control, and forceUpdate to avoid interfering with IME
+        if (!props.pair) {
+          if (event === 'input') {
+            doUpdateValue(targetValue, { source: index })
+          }
+          else {
+            doChange(targetValue, { source: index })
+          }
+        }
+        else {
+          let { value } = mergedValueRef
+          if (!Array.isArray(value)) {
+            value = ['', '']
+          }
+          else {
+            value = [value[0], value[1]]
+          }
+          value[index] = targetValue
+          if (event === 'input') {
+            doUpdateValue(value, { source: index })
+          }
+          else {
+            doChange(value, { source: index })
+          }
+        }
         return
+      }
       focusedInputCursorControl.recordCursor()
       const isIncomingValueValid = allowInput(targetValue)
       if (isIncomingValueValid) {


kyumoon:
I confirmed with Korean
스크린샷 2026-06-27 오후 2 03 31

kyumoon:
@Sepush @07akioni Could one of you take a look? Thanks!

MILLERMARRU (commented):
Traced `doUpdateValue`/`doChange` since the comment says this skips validation to avoid interfering with IME, and wanted to confirm that's actually true for everything, not just the input-pattern check. `doUpdateValue` unconditionally calls `nTriggerFormInput()`, which goes through `useFormItem` to `NFormItem.handleContentInput()`, which calls `internalValidate('input')`. That's real rule validation against the surrounding `n-form-item`, not just cursor/pattern handling, and this path isn't skipped by anything in this diff.

Before this PR, composition input never called `doUpdateValue`/`doChange` at all, the function returned immediately on `isComposingRef.value`, so form validation never ran mid-composition. After this PR, every intermediate `input` event during an active IME composition (which for CJK input can be several events before a character is even finalized) now triggers `doUpdateValue` → `nTriggerFormInput` → `internalValidate('input')` on the wrapping form item. So a field with an `input`-triggered validation rule would now show/update its validation state while the user is still composing a character, not just skip `allowInput`'s pattern check like the comment implies. That seems like exactly the kind of interference with IME the comment says this is meant to avoid, just from a different code path than the one it's guarding against.

The `doChange`/`compositionend` half is unaffected by this concern since that already ran full validation on commit before this PR too, this is specifically about the new mid-composition calls. Not saying the live-sync goal here is wrong, keeping the model in sync with the visible input while composing is a reasonable thing to want, just flagging that "skip validation" in the comment doesn't cover form-item rule validation the way it reads, and it'd be worth confirming whether that's intentional or whether `nTriggerFormInput`/`nTriggerFormChange` should also be gated on `!isComposingRef.value` the same way `allowInput` and cursor control are.

PR #8128, opened 2026-07-06
fix(date-picker): respect is-date-disabled in monthrange panel

### LINKED ISSUE #7513: n-date-picker使用monthrange时，isDateDisabled无效
### Describe the bug

`<n-date-picker
    v-model:value="filters.monthRange"
    type="monthrange"
    :is-date-disabled="isMonthDisabled"
  />`
type为monthrange时，isDateDisabled不会生效

### Steps to reproduce

月份范围组件设置isDateDisabled无效

### Link to minimal reproduction

https://codesandbox.io/p/sandbox/lucid-hooks-c7wsn2?file=%2Fsrc%2FDemo.vue

### System Info

```Shell
MacOS、chrome
```

### Used Package Manager

npm

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/tusen-ai/naive-ui/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://www.naiveui.com/en-US/).
- [x] Check that there isn't [already an issue](https://github.com/tusen-ai/naive-ui/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/tusen-ai/naive-ui/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

### LINKED ISSUE #8071: NDatePicker 设置type=‘monthrange’ 设置禁用属性无效
### Describe the bug

NDatePicker 设置type=‘monthrange’ 设置isDateDisabled无效 希望可以支持禁用

### Steps to reproduce

无

### Link to minimal reproduction

无

### System Info

```Shell
无
```

### Used Package Manager

npm

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/tusen-ai/naive-ui/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://www.naiveui.com/en-US/).
- [x] Check that there isn't [already an issue](https://github.com/tusen-ai/naive-ui/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/tusen-ai/naive-ui/discussions).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

lazerg:
The `monthrange` panel hard-coded `disabled` to `false` in `renderItem` (left as a `// TODO`), so `is-date-disabled` never took effect for `monthrange`, `quarterrange` or `yearrange`. This computes it from the panel's existing `mergedIsDateDisabled`, the same helper `daterange` already uses. Added a test for `monthrange`.

Fixes #7513, Fixes #8071

### DIFF
--- a/CHANGELOG.en-US.md
+++ b/CHANGELOG.en-US.md
@@ -35,6 +35,7 @@
 - Fix `n-date-picker` with `type="datetime"` emitting `update:value` twice when clearing from panel, closes [#8070](https://github.com/tusen-ai/naive-ui/issues/8070).
 - Fix `n-input-otp` only keeping the last character when browser extension autofill sets the full OTP value on a single input, closes [#7540](https://github.com/tusen-ai/naive-ui/pull/7540).
 - Fix `n-menu` item icons not centered in collapsed mode when the options contain a `type="group"` group, closes [#8105](https://github.com/tusen-ai/naive-ui/issues/8105).
+- Fix `n-date-picker` `is-date-disabled` not taking effect for `monthrange`, `quarterrange` and `yearrange`, closes [#7513](https://github.com/tusen-ai/naive-ui/issues/7513), [#8071](https://github.com/tusen-ai/naive-ui/issues/8071).
 
 ## 2.44.1
 
--- a/CHANGELOG.zh-CN.md
+++ b/CHANGELOG.zh-CN.md
@@ -35,6 +35,7 @@
 - 修复 `n-date-picker` 在 `type="datetime"` 时点击面板清空按钮会连续触发两次 `update:value` 的问题，关闭 [#8070](https://github.com/tusen-ai/naive-ui/issues/8070)
 - 修复 `n-input-otp` 在浏览器扩展自动填充时将完整 OTP 写入单个输入框时只保留最后一个字符的问题，关闭 [#7540](https://github.com/tusen-ai/naive-ui/pull/7540)
 - 修复 `n-menu` 在缩起（collapsed）状态下选项包含 `type="group"` 分组时菜单项图标未居中的问题，关闭 [#8105](https://github.com/tusen-ai/naive-ui/issues/8105)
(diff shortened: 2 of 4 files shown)


lazerg:
Hey @07akioni, could you take a look at this when you have time 🙏

MILLERMARRU (commented):
Checked `mergedIsDateDisabled` in `use-dual-calendar.ts` since that's what `useDualCalendar(props, props.type)` actually returns here, and it's a single-argument function, `(ts) => boolean`, that internally reads `isDateDisabledRef`, the current selecting phase and the memorized start time to decide range disabling on its own. So `mergedIsDateDisabled(item.ts)` is the complete, correct call, nothing extra needed. Worth noting for context since it could look incomplete at a glance, `use-calendar.ts` (used by the single date/month pickers) has a same-named function that takes a second `detail` argument, so if someone compares this against that one instead of the dual-calendar version it looks like a missing parameter. They're genuinely different functions for different picker families, so no issue here.

Confirmed `monthrange.tsx` backs `monthrange`, `quarterrange` and `yearrange` all through the same component via the `type` prop, which is why a one-file change closes both linked issues.

Test mounts with `isDateDisabled` forced to always return `true` and asserts every rendered picker column item actually got the `--disabled` class, which is a stronger check than just confirming the function got called, it verifies the class is threaded correctly from the disabled flag into the rendered node too, which is the part a wrong prop name or class list bug could still break even if `mergedIsDateDisabled` itself worked.

lazerg:
Thanks for the thorough check, glad the fix looks right to you.

lazerg:
This has been open a while with a confirmed-good review but no maintainer action yet. I'm closing it for now, feel free to reopen whenever it's convenient.

Issue #6481, opened 2024-10-21
naive-ui (2.40.1) menu 绑定主题 style 样式不生效


kn-msz:
### Describe the bug

![image](https://github.com/user-attachments/assets/bd33c8b6-f0f8-49ad-b162-260a709b0801)
![image](https://github.com/user-attachments/assets/a4dcd133-b803-4484-8182-00cd7a2863d7)

### Steps to reproduce

![image](https://github.com/user-attachments/assets/bd33c8b6-f0f8-49ad-b162-260a709b0801)
![image](https://github.com/user-attachments/assets/a4dcd133-b803-4484-8182-00cd7a2863d7)

### Link to minimal reproduction

un

### System Info

```Shell
1
```

### Used Package Manager

npm

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/tusen-ai/naive-ui/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://www.naiveui.com/en-US/).
- [X] Check that there isn't [already an issue](https://github.com/tusen-ai/naive-ui/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/tusen-ai/naive-ui/discussions).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

jahnli:
Provide a based on [CodeSandbox](https://codesandbox.io/templates/vue-vite) or [stackblitz](https://stackblitz.com/edit/vitejs-vite) `s [Minimal  Reproducible Example](https://stackoverflow.com/help/minimal-reproducible-example)

kn-msz:
codesandbox [https://codesandbox.io/p/devbox/naiveui-tree-p963zn](url)

> 提供基于[CodeSandbox](https://codesandbox.io/templates/vue-vite)或[stackblitz](https://stackblitz.com/edit/vitejs-vite)的[最小可重现示例](https://stackoverflow.com/help/minimal-reproducible-example)

kn-msz:
> Provide a based on [CodeSandbox](https://codesandbox.io/templates/vue-vite) or [stackblitz](https://stackblitz.com/edit/vitejs-vite) `s [Minimal Reproducible Example](https://stackoverflow.com/help/minimal-reproducible-example)

[...]

PR #8136, opened 2026-07-14
fix(menu): allow root style to override theme css vars

luozejian:
## 背景

修复 `NMenu` 根节点上传入的 `style` 无法覆盖主题 CSS 变量的问题。

以 `--n-item-text-color-active` 为例，用户通过组件根节点 `style` 传入自定义变量时，会被组件内部生成的默认 `cssVars` 覆盖，导致自定义主题样式不生效。

相关 issue：#6481

## 修改内容

- 调整 `src/menu/src/Menu.tsx` 中根节点属性合并逻辑
- 将根节点 `style` 改为合并组件内部 `cssVars` 和用户传入的 `attrStyle`
- 保证用户传入的根节点样式可以正确覆盖默认主题变量
- 保留原有根节点 `attrs` 透传能力，不影响 `id`、`class` 和普通内联样式

## 测试

在 `src/menu/tests/Menu.spec.tsx` 中补充了回归测试：

- 验证根节点 `style` 可以覆盖主题 CSS 变量
- 验证合并主题变量时不会丢失根节点 `id`、`class` 和普通 `style`

已通过以下验证：

- `npx pnpm@11.10.0 run test src/menu/tests/Menu.spec.tsx -t root`
- `npx pnpm@11.10.0 run test src/menu`
- `npx pnpm@11.10.0 exec eslint src/menu/src/Menu.tsx src/menu/tests/Menu.spec.tsx`


MILLERMARRU (commented):
Wanted to actually confirm why the old code lost, not just take the changeset's word for it, so I went into Vue core's `mergeProps` and `normalizeStyle`. `mergeProps` merges `style` as `normalizeStyle([ret.style, toMerge.style])`, and `normalizeStyle` flattens an array by iterating left to right and doing `res[key] = normalized[key]` for each item, so whichever style object appears later in the array wins for a shared key. In the old `mergeProps(this.$attrs, { style: this.cssVars })`, `this.$attrs` is processed first and `cssVars` second, so `cssVars` always landed later in the array and always won over anything the consumer passed in `style`. That's the exact opposite of what you'd want for a CSS-variable override.

Pulling `style` out of `$attrs` before calling `mergeProps` and building `style: [this.cssVars, attrStyle]` yourself is the right fix, since now `attrStyle` is deliberately placed last in the array that actually gets normalized, so it wins for any variable name it also sets, and `mergeProps` never gets a second `style` key from the spread `attrs` object to fight over ordering with.

Good that the second test checks `id`, `class`, and a plain `width` alongside the CSS variable, since it's the kind of change where it'd be easy to accidentally drop the rest of `$attrs` while pulling `style` out of it, and the destructure here doesn't.

luozejian:
@MILLERMARRU 
  Thanks for the detailed verification! That's exactly the intended behavior.
  I’ve synced the branch with the latest `main`; the implementation and
  regression tests remain unchanged.

PR #8140, opened 2026-07-16
fix(empty): support size via n-config-provider component-options

### LINKED ISSUE #8139: 【BUG】n-config-provider 无法控制 empty 组件 size
### 描述错误

n-config-provider 无法控制 empty 组件 size

### 复现步骤

可查看最小复现链接

### 最小复现链接

https://play-naive.pro-components.cn/#eNqtVE2P2jAQ/StWLoC0wKE9ReyqZcuhPbRVt8dIlTcZWO86dmQ7FEr5750ZBwgQaLXqzZ43bz6fvUneV9VoWUOSJpMAZaVlgLvMCDExw9yauVoMK2eXqgDHZga0lYUyi+GjdKco4yV4LxdwjjFqbFBzlcugrOlyiSFsIXU3ynihpLZntbVcPkBpxfiUOL7OJPxiZgKv18707t4Jujg2As+HPRm3FoJXnztVBeEh1JXQ0ixusyT4LEFUlZV1QXDPc2dL0RuN6UKb7VGkyEXP5CbZAV0rf6xDwM4WztbVYd/RKt7lWuUvmPVJmkLDNJh7MvR7QZl1b8CVxEGQ4dB3pP97OF9KrY/iseX1AUsoVF0eRYym14fU0i3gKCJbOgMerq3BsoZ51Y1Kzx+cSHOLizVgwtBWJDmPlextX6JpVwLycZlhLQqI20YQ3Yu1kaXK0YvKYA+WU5vh1S8gMdHWkhN+NP6dHNs/ZTfWTnozoaZAMeVrx+j+x8vYCBpbHaC4EQ7mYts8k/g+MhPWFYj73WQfsCVxK6KuxW/RKJJOjZToGCVAbKzQY/rIwvCTo0h3LQGi87w2OU1HnEjqB/HT4yIGqVhaVYgNzYLw0VLqmrKwd2a2h/SnukCnXc/9/kDc3ok+h4lzTjccL21H3d4QPqM1pTixcxzzDbCH/XeCn0nwcT2jZ28N/iicIopUaXB7jWLA+DayBEdpf35iW3A1cFLmPEH+0mF/9iuyZclXBx7cEvW0xwLtIER49vAZVnjeg/iT1xq9r4DfwFtdU43RbVrTTlzLj6v9yCrCv/u7n60CGL9rigolzzi6LEE90QIvtX4o983oLfNwpMn2Dx0egaY=

### 系统信息

```Shell
Naive UI Playground
```

### 使用的包管理器

npm

### 验证

seepine:
Fix #8139

n-config-provider 无法控制 empty 组件 size，因为：
1. `Empty` 类型在 `GlobalComponentConfig` 中没有声明 `size` 字段
2. Empty 组件没有从 `mergedComponentPropsRef` 中读取 size

## 修改

- `src/config-provider/src/internal-interface.ts`: 给 `Empty` 类型添加 `size` 字段
- `src/empty/src/Empty.tsx`:
  - 把 `size` 默认值从 `'medium'` 改为 `undefined`，以便正确检测用户是否传入了 size
  - 新增 `mergedSizeRef` 计算属性：props.size 优先，然后是 configSize，最后是 'medium'
  - `cssVarsRef` 和 themeClass hash 都使用 `mergedSizeRef` 而不是 `props.size`
- `src/empty/tests/Empty.spec.ts`: 新增测试用例验证 ConfigProvider 设置 size 的功能，以及 props.size 与 configSize 的优先级

### DIFF
--- a/CHANGELOG.en-US.md
+++ b/CHANGELOG.en-US.md
@@ -33,6 +33,7 @@
 - Fix `n-image` rendering the `placeholder` slot together with the `error` slot when the image fails to load, closes [#7526](https://github.com/tusen-ai/naive-ui/issues/7526).
 - Fix `n-space` `size` prop not taking effect when set to `0`, closes [#7530](https://github.com/tusen-ai/naive-ui/issues/7530).
 - Fix `n-date-picker` with `type="datetime"` emitting `update:value` twice when clearing from panel, closes [#8070](https://github.com/tusen-ai/naive-ui/issues/8070).
+- Fix `n-config-provider`'s `component-options` not controlling the `size` of `n-empty`, closes [#8139](https://github.com/tusen-ai/naive-ui/issues/8139).
 - Fix `n-input-otp` only keeping the last character when browser extension autofill sets the full OTP value on a single input, closes [#7540](https://github.com/tusen-ai/naive-ui/pull/7540).
 - Fix `n-menu` item icons not centered in collapsed mode when the options contain a `type="group"` group, closes [#8105](https://github.com/tusen-ai/naive-ui/issues/8105).
 
--- a/CHANGELOG.zh-CN.md
+++ b/CHANGELOG.zh-CN.md
@@ -33,6 +33,7 @@
 - 修复 `n-image` 图片加载失败时 `placeholder` 插槽与 `error` 插槽同时显示的问题，关闭 [#7526](https://github.com/tusen-ai/naive-ui/issues/7526)
 - 修复 `n-space` `size` 属性设置为 `0` 时不生效的问题，关闭 [#7530](https://github.com/tusen-ai/naive-ui/issues/7530)
 - 修复 `n-date-picker` 在 `type="datetime"` 时点击面板清空按钮会连续触发两次 `update:value` 的问题，关闭 [#8070](https://github.com/tusen-ai/naive-ui/issues/8070)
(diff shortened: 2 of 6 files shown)


MILLERMARRU (commented):
I just filed #8157 about this exact same bug class on Input/Select, so I reviewed this with that context still fresh and the fix here goes the right way.

The key part is `mergedSizeRef`. Switching the `size` prop default from `medium` to `undefined` is what actually lets you tell apart "user didn't pass size" from "user passed medium on purpose", and without that you can't implement the prop > component-options > default priority correctly at all. It's also the same pattern Button already uses internally, so this keeps things consistent instead of introducing a second way of resolving size.

On the reactivity side, `mergedComponentPropsRef?.value?.Empty?.size` read inside a computed tracks fine, optional chaining doesn't break Vue's dependency tracking since every intermediate get still goes through the reactive proxy until it short-circuits on null/undefined. The two new tests cover the cases that matter, component-options being picked up and the explicit prop winning when both are set.

Nothing to push back on logic-wise. One thing worth double checking is that `cssVarsRef` and the theme hash now read from `mergedSizeRef` instead of `props.size` directly, so it'd be worth a maintainer confirming nothing else in the file still reads `props.size` on its own and ends up out of sync. I went through the rest of Empty.tsx and didn't find one, but that's the kind of thing that's easy to miss in review.

PR #8151, opened 2026-08-06
fix(image): drop removed images from image group preview

### LINKED ISSUE #8150: n-image-group 的 n-image 被删除了，preview-prev 和 preview-next 还是能看到图片
### 描述错误

例如常见的图片墙场景，图片渲染后 image-group 能正常获取到src并且使用tool切换图片，但当有图片被删除，在tool切换图片中还是能看到
```ts
<n-image-group>
    <n-space>
      <div v-for="item of images" :key="item">
        <n-image width="100" :src="item" />
        <n-button @click="handleRemove(item)">删除</n-button>
      </div>
    </n-space>
  </n-image-group>
```

Image

### 复现步骤

参考 playground

### 最小复现链接

https://play-naive.pro-components.cn/#eNqVVMtu2zAQ/BVCF8dALDlogRSGbfSVAumhDZLeohwYciXTpkiCpGSnQe79hd566K/10L/oknJsOXEMFPDB2pldzs5Quk/eGZM2NSSjZOyhMpJ6mOaKkLEaMK0KUQ6M1Y3gYGM5AlJTLlQ5uKX2KRrxCpyjJTzHIqq0F4Vg1Aut9lHaEZpTuR+NOBdU6mfaOpSPUGmSPW3MDncG/MWTA3hYe2zfv3uAXrQtgM/NHmedQPDRMSuMJw58bYikqpzkiXd5gqiojLaexJ0LqyvSS7PwEJLthUltLzKT4+QR2Be5qIL20urabON2hrKWEQtcNKQZFNri+QIHEF2Q2IdayGgBd+t6VNaJLHLIUnA/Q8bJcBjozrINfScubLitvdeKvGVSsAWyZlRxCZcov4Gj0NLHI/78+PX35+/gYEvvhIFCO+5vtwhPu4v+v9X3xEJBHtZutzbnCkN0fm0GmQTK2HmLmV/fTI+uezPvjRtl2fCULvACiVQ7N2CoHMQcSSmV4q5WzKVMV1vS3EDZOw7CNwPKZeAaeqdv58B82/Bdu4wqXwjFuMroxeITe3M1P1VZBX6m+YBbumz3Tl1T9m76W8VdZ1F3NHdEWul9MpmS+3D8eju+Qkq7Y9pQWUOKR/JzxWF1hJ07fcKTyQTZMSwcIYoj7J9Ohv04EQvdOc5g0hgtXx2fRPpDrvC3e329a1+VdO60wjscB+UJOmCEBPvVhHcTsxq1ogNGpdTLz7HmbQ3Ry9gzA7bYU5+7VajlyYUFB7aBPNlgntoSfAufXX2BFf7fgPjtqCWyD4CX4LSsg8aW9r4O1tsOL6o9j9cMbfzmzlYelHtcKgiN1kR+nuDV+3Bg9a3cV+nrR0uTh39sNfWL

### 系统信息

```Shell
playground
```

### 使用的包管理器

npm

### 验证

lazerg:
When an `n-image` inside an `n-image-group` was removed, its entry stayed in the group's registered url map, so preview-prev and preview-next could still switch to the deleted image. The unregister callback had its condition inverted (`if (!map.has(sid))`), so it only ran when the entry was already gone.

The callback now deletes the entry when it is there, and `n-image` calls it on unmount instead of on every effect invalidation, so updating a `src` still replaces the url in place and keeps the image's position in the group.

Fixes #8150

### DIFF
--- a/CHANGELOG.en-US.md
+++ b/CHANGELOG.en-US.md
@@ -35,6 +35,7 @@
 - Fix `n-date-picker` with `type="datetime"` emitting `update:value` twice when clearing from panel, closes [#8070](https://github.com/tusen-ai/naive-ui/issues/8070).
 - Fix `n-input-otp` only keeping the last character when browser extension autofill sets the full OTP value on a single input, closes [#7540](https://github.com/tusen-ai/naive-ui/pull/7540).
 - Fix `n-menu` item icons not centered in collapsed mode when the options contain a `type="group"` group, closes [#8105](https://github.com/tusen-ai/naive-ui/issues/8105).
+- Fix `n-image-group` still previewing images whose `n-image` has been removed, closes [#8150](https://github.com/tusen-ai/naive-ui/issues/8150).
 
 ## 2.44.1
 
--- a/CHANGELOG.zh-CN.md
+++ b/CHANGELOG.zh-CN.md
@@ -35,6 +35,7 @@
 - 修复 `n-date-picker` 在 `type="datetime"` 时点击面板清空按钮会连续触发两次 `update:value` 的问题，关闭 [#8070](https://github.com/tusen-ai/naive-ui/issues/8070)
 - 修复 `n-input-otp` 在浏览器扩展自动填充时将完整 OTP 写入单个输入框时只保留最后一个字符的问题，关闭 [#7540](https://github.com/tusen-ai/naive-ui/pull/7540)
 - 修复 `n-menu` 在缩起（collapsed）状态下选项包含 `type="group"` 分组时菜单项图标未居中的问题，关闭 [#8105](https://github.com/tusen-ai/naive-ui/issues/8105)
+- 修复 `n-image-group` 中的 `n-image` 被移除后仍可在预览中切换到该图片的问题，关闭 [#8150](https://github.com/tusen-ai/naive-ui/issues/8150)
 
 ## 2.44.1
 
--- a/src/image/src/Image.tsx
+++ b/src/image/src/Image.tsx
@@ -126,14 +126,15 @@ export default defineComponent({
       showErrorRef.value = false
     })
 
-    watchEffect((onInvalidate) => {
-      const unregister = imageGroupHandle?.registerImageUrl?.(
+    let unregisterImageUrl: (() => void) | undefined
+    watchEffect(() => {
+      unregisterImageUrl = imageGroupHandle?.registerImageUrl?.(
         imageId,
         mergedPreviewSrcRef.value || ''
       )
-      onInvalidate(() => {
-        unregister?.()
-      })
+    })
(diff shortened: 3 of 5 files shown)


lazerg:
The failing test job is the n-space snapshot mismatch. It fails the same way on main without this change, so it does not come from this PR.

MILLERMARRU (commented):
The real fix here is the one-line flip in `ImageGroup.tsx`, `if (!registeredImageUrlMap.value.has(sid))` becoming `if (registeredImageUrlMap.value.has(sid))`. That inverted check is exactly why #8150 happened: `unregisterPreviewUrl` only ran `delete` when the key was already absent, so it was a no-op every single time an `n-image` unmounted, and the stale sid stayed in `registeredImageUrlMap` forever, which is why the group preview could still land on an image that was already removed from the DOM. Good catch, that's the actual root cause and the fix is correct.

While reading the surrounding code I noticed `registerImageUrl` has basically the same class of bug on the other side, `if (!registeredImageUrlMap.value.has(\`r${sid}\`))` where `sid` is already `\`r${id}\``, so it's checking for a key like `rr123` that can never exist. That guard is dead code, it always evaluates true, so registering always overwrites. It happens to be harmless right now because `imageId` is a stable per-instance uuid set once in `Image.tsx` setup, so the key never actually collides in a way that guard was meant to prevent, but it's the same "wrong key in the has() check" mistake and probably deserves its own fix even if it's out of scope for this PR.

On the `Image.tsx` side, moving from `watchEffect`'s `onInvalidate` to a plain closure variable plus `onBeforeUnmount` works fine today because `imageId` never changes for the component's lifetime, so the effect only re-runs when `src` changes and re-registering under the same `sid` just overwrites the map entry in place, nothing needs to be unregistered first. Just flagging that this relies on `imageId` staying non-reactive. If that ever changes, this version would leak the previous registration since the old `unregisterImageUrl` gets overwritten without being called, where the old `onInvalidate` version would have cleaned it up automatically before each re-run.

The two new tests are the right ones to add, one for the removed-image case from the issue and one that specifically checks index/order stays correct when a src is swapped rather than removed, which is a good sign the fix was checked against a case that could have regressed.

lazerg:
Thanks for confirming the fix. Good catch on the `registerImageUrl` guard. It's already filed as #8158, and #8168 has an open fix for it, so I won't duplicate it here. On the `onBeforeUnmount` point: agreed, it relies on `imageId` staying non-reactive. That holds today, so I'm leaving it as is for now.

lazerg:
Hey @07akioni, could you take a look at this fix that drops removed images from the image group preview? MILLERMARRU already reviewed it, it just needs a maintainer call.

lazerg:

PR #8155, opened 2026-08-07
fix(data-table): escape line breaks and quotes in exported CSV

### LINKED ISSUE #8153: 【Data Table】列的文字有换行，导出 CSV 格式不正确
### 描述错误

Image

数据有个换行，展示也没问题。
导出 CSV 的列文案换行到第二行第一列了。

Image

### 复现步骤

本来想在 playground 复现，但是它点击导出没效果，后来跑去 CodeSandbox，它编译报错了：

Image

Image

只能麻烦开发者前往 https://github.com/sunpm/naiveui-issue/tree/2026-08-07 拉取编译复现了

### 最小复现链接

https://github.com/sunpm/naiveui-issue/tree/2026-08-07

### 系统信息

```Shell
https://github.com/sunpm/naiveui-issue/tree/2026-08-07
```

### 使用的包管理器

pnpm

### 验证

lazerg:
`formatCsvCell` only escaped commas, and it did so with a backslash (`a\,b`), which no CSV parser un-escapes. A cell that holds a line break, a comma or a double quote therefore leaks into the next row or column when the table is exported, as the reporter shows with an address value containing `\r\n`. Column titles were not escaped at all, so a title with a comma splits the header row the same way.

Cells and default column titles are now wrapped in double quotes when they contain `,`, `"`, `\r` or `\n`, with inner quotes doubled, following RFC 4180. Plain values stay unquoted. Output from `get-csv-cell` and `get-csv-header` is still written as given, since those already return the final text.

Fixes #8153

### DIFF
--- a/CHANGELOG.en-US.md
+++ b/CHANGELOG.en-US.md
@@ -35,6 +35,7 @@
 - Fix `n-date-picker` with `type="datetime"` emitting `update:value` twice when clearing from panel, closes [#8070](https://github.com/tusen-ai/naive-ui/issues/8070).
 - Fix `n-input-otp` only keeping the last character when browser extension autofill sets the full OTP value on a single input, closes [#7540](https://github.com/tusen-ai/naive-ui/pull/7540).
 - Fix `n-menu` item icons not centered in collapsed mode when the options contain a `type="group"` group, closes [#8105](https://github.com/tusen-ai/naive-ui/issues/8105).
+- Fix `n-data-table`'s exported CSV being broken by cells containing line breaks, commas or double quotes, closes [#8153](https://github.com/tusen-ai/naive-ui/issues/8153).
 
 ## 2.44.1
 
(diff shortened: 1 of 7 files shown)


lazerg:
The failing test job is the n-space, n-time-picker and locale snapshot mismatch that main already has, not something this change causes. The data-table tests pass.

MILLERMARRU (commented):
This is a real correctness fix, not just a style change. The old `formatCsvCell` only escaped commas by prefixing them with a backslash, which isn't valid CSV syntax at all, so any consumer parsing the export with a real CSV parser (Excel, `csv-parse`, pandas) would either choke on it or read the row wrong. Line breaks inside a cell weren't handled at all before, so a multiline value would silently split into extra rows on import.

Switching to RFC 4180 style quoting (wrap in double quotes when the value has a comma, quote or CR/LF, and double up internal quotes) is the correct fix and matches what basically every CSV reader expects. The regex `/["\,\r\n]/` covers all three trigger cases and the doubled `""` escaping is right.

Applying `formatCsvCell` to `col.title` too is a good catch since headers can obviously contain commas as well, and that path was completely unescaped before.

One thing I noticed reading the surrounding code: if a caller passes a custom `getCsvCell` or `getCsvHeader`, that output still bypasses `formatCsvCell` entirely, same as before this PR. Not a regression since that's pre-existing behavior and arguably the caller's responsibility to escape their own formatter output, just flagging it in case it's worth a follow-up note in the docs so people don't assume custom formatters get auto-escaped.

Tests look solid, they cover the plain case, comma, quote, line break, header quoting and null/undefined all separately instead of one big combined test, which makes future regressions easy to pinpoint.

lazerg:
Good catch, thanks. You're right that it wasn't documented anywhere. Added a note to both `getCsvCell` and `getCsvHeader` (JSDoc on the type plus the props table in both docs locales) stating the returned string is used as-is and callers must escape commas/quotes/line breaks themselves: 91ef5af57.

lazerg:
Hey @07akioni, could you review this fix that escapes line breaks and quotes in exported CSV? MILLERMARRU already reviewed it, just needs a maintainer decision.

lazerg:

PR #8162, opened 2026-08-09
fix(avatar): read global size config from n-config-provider

### LINKED ISSUE #8160: n-config-provider's component-options doesn't work for Avatar's size
### Describe the bug

`n-config-provider`'s `component-options` prop doesn't work for `Avatar`'s `size`, even though `Avatar` has the `useConfig` infrastructure available to support it, it just never wires it up.

In `src/avatar/src/Avatar.tsx`:

```ts
const { mergedClsPrefixRef, inlineThemeDisabled } = useConfig(props)
// ...
const mergedSizeRef = computed(() => {
  const { size } = props
  if (size)
    return size
  const { size: avatarGroupSize } = NAvatarGroup || {}
  if (avatarGroupSize)
    return avatarGroupSize
  return 'medium'
})
```

`useConfig(props)` is called, but only `mergedClsPrefixRef` and `inlineThemeDisabled` are destructured from it, `mergedComponentPropsRef` is never pulled out or checked. `mergedSizeRef` only considers the component's own `size` prop and the surrounding `n-avatar-group`, so `component-options: { Avatar: { size: 'large' } }` has no effect.

I confirmed this is a real, working mechanism elsewhere by checking `Button.tsx`, which resolves its size the same way except it adds one more fallback step, `mergedComponentPropsRef?.value?.Button?.size`, before defaulting to `'medium'`.

### Steps to reproduce

1. Wrap an `n-avatar` (without its own `size` prop and outside an `n-avatar-group`) in `<n-config-provider :component-options="{ Avatar: { size: 'large' } }">`.
2. Observe the avatar renders at the default `medium` size instead of `large`.

### Expected behavior

`Avatar` should check `mergedComponentPropsRef?.value?.Avatar?.size` as a fallback before its hardcoded `'medium'` default, matching the pattern already used by `Button`.

### System Info

Not applicable, this is a source-level gap (an unused piece of `useConfig`'s return value) rather than an environment-dependent bug.

lazerg:
`n-avatar` calls `useConfig` but never destructures `mergedComponentPropsRef` from it, so a `size` set through `n-config-provider`'s `component-options` was ignored and the avatar always fell back to `medium`.

`mergedSizeRef` now checks that config, after the component's own `size` prop and the surrounding `n-avatar-group`, the same order `n-button` uses. `Avatar` is added to `GlobalComponentConfig` and to the existing `ComponentSize` spec.

Fixes #8160

### DIFF
--- a/CHANGELOG.en-US.md
+++ b/CHANGELOG.en-US.md
@@ -35,6 +35,7 @@
 - Fix `n-date-picker` with `type="datetime"` emitting `update:value` twice when clearing from panel, closes [#8070](https://github.com/tusen-ai/naive-ui/issues/8070).
 - Fix `n-input-otp` only keeping the last character when browser extension autofill sets the full OTP value on a single input, closes [#7540](https://github.com/tusen-ai/naive-ui/pull/7540).
 - Fix `n-menu` item icons not centered in collapsed mode when the options contain a `type="group"` group, closes [#8105](https://github.com/tusen-ai/naive-ui/issues/8105).
+- Fix `n-avatar` not reading global `size` config from `n-config-provider`, closes [#8160](https://github.com/tusen-ai/naive-ui/issues/8160).
 
 ## 2.44.1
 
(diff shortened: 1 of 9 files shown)


lazerg:
The failing `n-tabs` "should work with `size` prop" test also fails on a clean `main` checkout, so it does not come from this change.

MILLERMARRU (commented):
I filed #8160 this fix closes, so happy to confirm it's the right shape. `mergedComponentPropsRef?.value?.Avatar?.size` slotted in between the avatar-group size and the hardcoded `'medium'` fallback matches the precedence every other sized component already uses, checked it against Button's `mergedSizeRef` (own prop, then contextual override, then `mergedComponentPropsRef?.value?.Button?.size`, then default), same order here with avatar-group standing in for the contextual layer.

The `public-types.ts` + `AvatarSize` re-export also follows the existing convention other components use for their config-provider entry (matches what's already done for `AutoCompleteSize` right above it in `internal-interface.ts`), so this isn't introducing a new pattern, just extending the one already in place to a component that was missing it. Test covers the actual regression, `n-config-provider` setting a global Avatar size and the component picking it up.

lazerg:
Closing this since the linked issue #8160 has been closed already, so this fix isn't needed anymore.

PR #8168, opened 2026-08-09
fix(image): correct duplicate registration guard

### LINKED ISSUE #8158: registerImageUrl's re-registration guard checks the wrong map key and never actually short-circuits
### Describe the bug

`registerImageUrl` in `src/image/src/ImageGroup.tsx` builds its map key as `sid = \`r${id}\``, then guards the initial `set` with:

```ts
const sid = `r${id}`
if (!registeredImageUrlMap.value.has(`r${sid}`)) {
  registeredImageUrlMap.value.set(sid, url)
}
```

The `has()` check looks up `` `r${sid}` `` (i.e. `` `rr${id}` ``), not `sid` itself. Entries are always stored under `sid` (`` `r${id}` ``), so a key like `` `rr${id}` `` can never exist in the map, the condition is always `true`, and the "already registered" guard never actually short-circuits, it always falls through to `set`.

This is the sibling of the bug just fixed on the unregister side in #8151 (an inverted `has()` check there left removed images registered forever, tracked in #8150). Same root cause, wrong key passed to `has()`, just on the registration path instead of the cleanup path.

It doesn't cause a visible runtime bug today because `imageId` is a stable per-component-instance value (`let uuid = 0; const imageId = uuid++` in `Image.tsx`), so re-registering under the same `sid` just overwrites the same entry with the same effective result whether the guard runs or not. But the guard clearly exists to make repeated registration calls a no-op, and right now it silently never does that.

### Steps to reproduce

Static logic error, not runtime-reproducible as visible breakage today, verifiable by reading `registerImageUrl` in `src/image/src/ImageGroup.tsx` directly and comparing the key built (`sid`) against the key checked (`` `r${sid}` ``) in the `has()` call two lines below it.

### Link to minimal reproduction

Found while reviewing #8151, which fixes the corresponding bug on `unregisterPreviewUrl` (tracked in #8150) but doesn't touch this line.

### System Info

Not applicable, this is a static logic error (wrong key in a `Map.has()` call) rather than an environment-dependent runtime bug.

jahnli:
close https://github.com/tusen-ai/naive-ui/issues/8158

### DIFF
--- a/CHANGELOG.en-US.md
+++ b/CHANGELOG.en-US.md
@@ -35,6 +35,7 @@
 - Fix `n-date-picker` with `type="datetime"` emitting `update:value` twice when clearing from panel, closes [#8070](https://github.com/tusen-ai/naive-ui/issues/8070).
 - Fix `n-input-otp` only keeping the last character when browser extension autofill sets the full OTP value on a single input, closes [#7540](https://github.com/tusen-ai/naive-ui/pull/7540).
 - Fix `n-menu` item icons not centered in collapsed mode when the options contain a `type="group"` group, closes [#8105](https://github.com/tusen-ai/naive-ui/issues/8105).
+- Fix `n-image-group` registration guard using the wrong map key for duplicate image registration, closes [#8158](https://github.com/tusen-ai/naive-ui/issues/8158).
 
 ## 2.44.1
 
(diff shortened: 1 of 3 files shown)


MILLERMARRU (commented):
The fix itself is correct, `has(sid)` matches the actual key used by `.set(sid, url)` a few lines down, the old `has(\`r${sid}\`)` was checking for a key with the prefix applied twice, which is never the key actually stored, so the guard could never trigger and the branch always ran.

While confirming that, I noticed `unregisterPreviewUrl` a few lines below has what looks like the same class of bug, inverted rather than doubled:

```ts
return function unregisterPreviewUrl() {
  if (!registeredImageUrlMap.value.has(sid)) {
    registeredImageUrlMap.value.delete(sid)
  }
}
```

`Map.delete()` is a no-op on a missing key, so this only calls `delete()` in exactly the case where it does nothing (key already absent), and skips it in the case where the key is actually present, i.e. the normal unmount path. Ran it standalone to confirm the semantics aren't doing something non-obvious:

```
has r1: true
after guarded delete attempt, still has r1: true
```

So as written, an `n-image` inside an `n-image-group` never actually gets removed from `registeredImageUrlMap` on unmount. Given this PR is already touching the surrounding function for the registration-guard fix, might be worth flipping this one too (`if (registeredImageUrlMap.value.has(sid))`) in the same change rather than leaving it for a separate issue.

PR #8174, opened 2026-08-13
fix(tree-select): ignore Enter while IME composition is active

### LINKED ISSUE #8173: n-tree-select: Enter confirms a filtered node during IME composition (missing isComposing guard)
### Description

Same bug class as #8115 and #8172 (which I just filed for `n-cascader`), and a third component sharing it: `n-tree-select`. Its `handleKeydown` also confirms the currently keyboard-focused (filtered) tree node on `Enter` with no IME composition guard:

```ts
function handleKeydown(e: KeyboardEvent): void {
  if (e.key === 'Enter') {
    if (mergedShowRef.value) {
      const { enterBehavior } = treeHandleKeydown(e)
      if (!props.multiple) {
        switch (enterBehavior) {
          case 'default':
          case 'toggleSelect':
            closeMenu()
            focusSelection()
            break
          // ...
```

Like Cascader, TreeSelect renders its trigger via `NInternalSelection` (same component `n-select`/`n-cascader` use), which already exposes `isComposing`, TreeSelect just never checks it before acting on `Enter`.

### Reproduction

1. Render a filterable `n-tree-select`.
2. Open the menu and start typing a filter value using an IME.
3. Press `Enter` to confirm the IME composition candidate while composing.

Expected: composition confirms, no tree node gets selected yet. Actual: the currently keyboard-focused (filtered) tree node gets toggled/selected immediately from the half-composed filter text, and (in single-select mode) the menu closes.

### Suggested fix

Same shape as the Cascader fix:

```ts
function handleKeydown(e: KeyboardEvent): void {
  if (e.key === 'Enter') {
    if (this.triggerInstRef?.isComposing) return
    if (mergedShowRef.value) {
      // ...
```

(again, exact property access may need adjusting to however `triggerInstRef` is exposed in this file, but `isComposing` already exists on the underlying `NInternalSelection` instance)

lazerg:
`n-tree-select`'s `handleKeydown` acts on `Enter` without checking whether an IME composition is in progress, so the Enter that confirms a conversion candidate also toggles the currently keyboard-focused tree node and closes the menu in single-select mode.

The trigger is `NInternalSelection`, which already tracks `isComposing`, so the fix is the same guard `n-select` uses: skip the Enter handling while composing, and still call `preventDefault`.

Fixes #8173

### DIFF
--- a/CHANGELOG.en-US.md
+++ b/CHANGELOG.en-US.md
@@ -45,6 +45,7 @@
 - Fix `n-input-otp` only keeping the last character when browser extension autofill sets the full OTP value on a single input, closes [#7540](https://github.com/tusen-ai/naive-ui/pull/7540).
 - Fix `n-menu` item icons not centered in collapsed mode when the options contain a `type="group"` group, closes [#8105](https://github.com/tusen-ai/naive-ui/issues/8105).
 - Fix `n-data-table` not rendering the table header when data is empty and a column has `ellipsis` set, closes [#8118](https://github.com/tusen-ai/naive-ui/issues/8118).
+- Fix `n-tree-select` selecting the keyboard-focused node when Enter is pressed during IME composition, closes [#8173](https://github.com/tusen-ai/naive-ui/issues/8173).
 
 ## 2.44.1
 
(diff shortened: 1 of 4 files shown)


lazerg:

PR #8175, opened 2026-08-13
fix(cascader): ignore Enter while IME composition is active

### LINKED ISSUE #8172: n-cascader: Enter confirms a filtered option during IME composition (missing isComposing guard)
### Description

Same bug class as #8115 (`n-dynamic-tags`), just in a different component. `n-cascader`'s `handleKeydown` confirms the currently highlighted filtered option on `Enter` with no IME composition guard at all:

```ts
function handleKeydown(e: KeyboardEvent): void {
  switch (e.key) {
    // ...
    case 'Enter':
      if (!mergedShowRef.value) {
        openMenu()
      }
      else {
        // ...
        else {
          if (selectMenuInstRef.value) {
            const hasCorrespondingOption = selectMenuInstRef.value.enter()
            if (hasCorrespondingOption)
              clearPattern()
          }
        }
      }
      break
```

There's no `isComposing`/`isCompositing` check anywhere in `Cascader.tsx` (I grepped the whole file). When `filterable` is set and a user is typing via an IME (Japanese/Chinese/Korean) to filter the option list, the `Enter` that's meant to confirm a composition candidate still reaches this handler while composing, and since the filtered menu is open, it immediately confirms whatever cascader option happens to be highlighted by the half-typed filter text instead of letting the IME finish composing.

`n-select` already guards this exact case in the same situation:

```ts
case 'Enter':
  if (!triggerRef.value?.isComposing) {
    // ...
```

And Cascader uses the same underlying trigger component (`NInternalSelection`, bound via `ref="triggerInstRef"` in the render function) that Select reads `isComposing` from, so the same property should be available here, `n-cascader` just never checks it.

### Reproduction

1. Render a filterable `n-cascader`.
2. Open the menu and start typing a filter value using an IME (any environment with a CJK input method works, or simulate by dispatching `compositionstart` before the `Enter` keydown, then `compositionend` after).
3. Press `Enter` to confirm the IME composition candidate while composing.

Expected: the composition confirms, no cascader selection happens yet. Actual: the currently highlighted (filtered) cascader option gets confirmed immediately, using whatever partial/incorrect filter text was present mid-composition.

### Suggested fix

```ts
case 'Enter':
  if (this.triggerInstRef?.isComposing) return
  if (!mergedShowRef.value) {
    openMenu()
  }
  // ...
```

(exact property access may need adjusting depending on how `triggerInstRef` is typed/exposed in this file, but the property itself already exists on `NInternalSelection`)

lazerg:
When a filterable `n-cascader` is filtered with an IME, the Enter that confirms the conversion candidate still reaches the cascader's keydown handler, so the highlighted option gets selected before the user has finished typing.

`n-select` already skips Enter while the trigger reports `isComposing`, and cascader uses the same `NInternalSelection` trigger, so it can read the same flag. Added a spec covering the composing and non-composing Enter, plus changelog entries.

Fixes #8172

### DIFF
--- a/CHANGELOG.en-US.md
+++ b/CHANGELOG.en-US.md
@@ -45,6 +45,7 @@
 - Fix `n-input-otp` only keeping the last character when browser extension autofill sets the full OTP value on a single input, closes [#7540](https://github.com/tusen-ai/naive-ui/pull/7540).
 - Fix `n-menu` item icons not centered in collapsed mode when the options contain a `type="group"` group, closes [#8105](https://github.com/tusen-ai/naive-ui/issues/8105).
 - Fix `n-data-table` not rendering the table header when data is empty and a column has `ellipsis` set, closes [#8118](https://github.com/tusen-ai/naive-ui/issues/8118).
+- Fix `n-cascader` selecting the highlighted option when Enter is pressed to confirm an IME composition, closes [#8172](https://github.com/tusen-ai/naive-ui/issues/8172).
 
 ## 2.44.1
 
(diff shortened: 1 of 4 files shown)


lazerg:

PR #8177, opened 2026-08-14
fix(image): reset loaded state when src changes

### LINKED ISSUE #8176: n-image: placeholder never reappears after src changes because loaded is never reset (mirror gap of #8123)
I read #8123 ("Fix `n-image` rendering the `placeholder` slot together with the `error` slot when the image fails to load", closes #7526), which added `loadedRef.value = true` inside `mergedOnError` in `src/image/src/Image.tsx` so the `error` slot correctly replaces the `placeholder` slot once the image fails, instead of both showing at once.

I checked the other `watchEffect` in the same file that reacts to the image `src` changing:

```ts
watchEffect(() => {
  void (props.src || props.imgProps?.src)
  showErrorRef.value = false
})
```

This resets `showErrorRef` whenever `src` (or `imgProps.src`) changes, but it never resets `loadedRef`. `loadedRef` is only ever set to `true` (in `mergedOnLoad` and, since #8123, in `mergedOnError`), there's no code path that sets it back to `false`.

In `render()`, both the placeholder visibility and the ``'s own hidden styling key off `loaded`:

```tsx
style: [
  ...,
  placeholderNode && !loaded
    ? { height: '0', width: '0', visibility: 'hidden' }
    : '',
  ...
]
...
{!loaded && placeholderNode}
```

So the sequence is: an `n-image` loads successfully once (`loadedRef` becomes `true`), then the consumer changes `src` to point at a different image. `showErrorRef` correctly resets to `false`, but `loadedRef` stays `true` from the previous load, so:
- the placeholder slot never reappears while the new image is loading (`!loaded` is `false`)
- the `` element is never hidden/collapsed while the new source is loading, so the old image (or a broken-image icon, depending on load timing) stays visible at full size instead of yielding to the placeholder

This is the mirror-image gap of what #8123 fixed: that PR made sure `loaded` becomes `true` on error so the error slot can take over from the placeholder; this case needs `loaded` to go back to `false` when `src` changes so the placeholder can take over again for the new load.

**Suggested fix**, in the same `watchEffect` #8123 didn't touch:

```ts
watchEffect(() => {
  void (props.src || props.imgProps?.src)
  showErrorRef.value = false
  loadedRef.value = false
})
```

(`loadedRef` would need to be declared before this `watchEffect` instead of further down in `setup()`, since it currently isn't in scope at that point.)

Checked against the current `main` source of `src/image/src/Image.tsx`.

lazerg:
Once an image loaded, `loaded` stayed true forever, so changing `src` left the old image on screen and the placeholder slot never came back while the new source was loading. The watchEffect that already resets `showError` on a `src` change now resets `loaded` too.

Fixes #8176

### DIFF
--- a/CHANGELOG.en-US.md
+++ b/CHANGELOG.en-US.md
@@ -45,6 +45,7 @@
 - Fix `n-input-otp` only keeping the last character when browser extension autofill sets the full OTP value on a single input, closes [#7540](https://github.com/tusen-ai/naive-ui/pull/7540).
 - Fix `n-menu` item icons not centered in collapsed mode when the options contain a `type="group"` group, closes [#8105](https://github.com/tusen-ai/naive-ui/issues/8105).
 - Fix `n-data-table` not rendering the table header when data is empty and a column has `ellipsis` set, closes [#8118](https://github.com/tusen-ai/naive-ui/issues/8118).
+- Fix `n-image` not showing the `placeholder` slot again when `src` changes, closes [#8176](https://github.com/tusen-ai/naive-ui/issues/8176).
 
 ## 2.44.1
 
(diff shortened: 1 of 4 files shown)


lazerg:

PR #8181, opened 2026-08-15
fix(avatar): reset loaded state when src changes

### LINKED ISSUE #8180: n-avatar: placeholder never reappears after src changes because loadedRef is never reset
This is the same gap I found for n-image in #8176 (mirror of #8123), just in `Avatar.tsx` instead.

Avatar resets `hasLoadErrorRef` when `src` changes but never resets `loadedRef`:

```js
watch(
  () => props.src || props.imgProps?.src,
  () => { hasLoadErrorRef.value = false },
)

const loadedRef = ref(!props.lazy)
```

`loadedRef` is only ever set to `true`, inside `mergedOnLoad`. There's no path that sets it back to `false`.

`render()` gates both the placeholder and the ``'s own hidden styling on `loaded`:

```js
const placeholderNode = !loaded && !hasLoadError && (...)
...
placeholderNode ? { height: '0', width: '0', visibility: 'hidden', position: 'absolute' } : ''
```

So: an `n-avatar` loads `src` A successfully (`loadedRef` becomes `true`), then the consumer changes `src` to B. `hasLoadErrorRef` correctly resets, `loadedRef` stays `true` from the previous load, so the placeholder slot never reappears while the new image is loading (`!loaded` is `false`), and the `` is never hidden/collapsed while the new source loads, so the stale image (or a broken-image icon) stays visible at full size instead of yielding to the placeholder.

Suggested fix, same watch that already resets `hasLoadErrorRef`:

```js
watch(
  () => props.src || props.imgProps?.src,
  () => {
    hasLoadErrorRef.value = false
    loadedRef.value = false
  },
)
```

(`loadedRef` would need to be declared before this watch instead of further down in `setup()`, same note as in #8176.)

Confirmed against the current `main` source of `src/avatar/src/Avatar.tsx`.

lazerg:
`loadedRef` in `Avatar.tsx` was only ever set to `true`, in `mergedOnLoad`. The watch that already resets `hasLoadErrorRef` when `src` changes left it alone, so after one successful load the placeholder slot never came back and the `` was never collapsed while the next source loaded.

The watch now resets it too, back to `!props.lazy`, which is the same value the ref is initialized with, so a non-lazy avatar keeps behaving as before.

Fixes #8180

### DIFF
--- a/CHANGELOG.en-US.md
+++ b/CHANGELOG.en-US.md
@@ -45,6 +45,7 @@
 - Fix `n-input-otp` only keeping the last character when browser extension autofill sets the full OTP value on a single input, closes [#7540](https://github.com/tusen-ai/naive-ui/pull/7540).
 - Fix `n-menu` item icons not centered in collapsed mode when the options contain a `type="group"` group, closes [#8105](https://github.com/tusen-ai/naive-ui/issues/8105).
 - Fix `n-data-table` not rendering the table header when data is empty and a column has `ellipsis` set, closes [#8118](https://github.com/tusen-ai/naive-ui/issues/8118).
+- Fix `n-avatar` not showing the `placeholder` slot again when `src` changes, closes [#8180](https://github.com/tusen-ai/naive-ui/issues/8180).
 
 ## 2.44.1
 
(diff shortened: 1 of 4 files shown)


lazerg:

PR #8182, opened 2026-08-15
fix(input-number): ignore Enter while IME composition is active

### LINKED ISSUE #8179: n-input-number: Enter commits the value during IME composition (missing isCompositing guard)
InputNumber's `handleKeyDown` commits on Enter without checking whether an IME composition is active:

```js
function handleKeyDown(e: KeyboardEvent): void {
  if (e.key === 'Enter') {
    if (e.target === inputInstRef.value?.wrapperElRef) return
    const value = deriveValueFromDisplayedValue({ offset: 0, doUpdateIfValid: true, isInputing: false, fixPrecision: true })
    if (value !== false) inputInstRef.value?.deactivate()
  }
  ...
}
```
(`src/input-number/src/InputNumber.tsx`)

`inputInstRef` is the underlying `NInput`, and `NInput` already exposes an `isCompositing` flag for exactly this case (`src/input/src/Input.tsx`, `isCompositing: isComposingRef`). `n-select` already checks it before treating Enter as a selection (`triggerRef.value?.isComposing` in `src/select/src/Select.tsx`), and I filed the same class of gap for `n-cascader`/`n-tree-select` a few rounds back (#8172, #8173). InputNumber never references composition state at all, I grepped the whole file for "composi" and got zero matches.

This matters for full-width digit input through CJK IMEs: pressing Enter to confirm a composition candidate reaches this handler while composing, so it derives and commits the not-yet-finalized displayed value and calls `deactivate()` on the input, swallowing the Enter the IME needed to finalize the candidate instead of letting composition complete normally.

Fix mirrors Select's:

```js
function handleKeyDown(e: KeyboardEvent): void {
  if (e.key === 'Enter') {
    if (inputInstRef.value?.isCompositing) return
    ...
```

lazerg:
`n-input-number`'s `handleKeyDown` acted on Enter without looking at composition state, so the Enter that confirms an IME candidate also derived and committed the displayed value and deactivated the input.

The underlying `n-input` already exposes `isCompositing`, so the Enter branch now returns early while a composition is active, the same way `n-select` and `n-dynamic-tags` do.

Fixes #8179

### DIFF
--- a/CHANGELOG.en-US.md
+++ b/CHANGELOG.en-US.md
@@ -45,6 +45,7 @@
 - Fix `n-input-otp` only keeping the last character when browser extension autofill sets the full OTP value on a single input, closes [#7540](https://github.com/tusen-ai/naive-ui/pull/7540).
 - Fix `n-menu` item icons not centered in collapsed mode when the options contain a `type="group"` group, closes [#8105](https://github.com/tusen-ai/naive-ui/issues/8105).
 - Fix `n-data-table` not rendering the table header when data is empty and a column has `ellipsis` set, closes [#8118](https://github.com/tusen-ai/naive-ui/issues/8118).
+- Fix `n-input-number` committing the value when Enter is pressed to confirm an IME composition, closes [#8179](https://github.com/tusen-ai/naive-ui/issues/8179).
 
 ## 2.44.1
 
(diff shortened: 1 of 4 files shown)


lazerg:

PR #8205, opened 2026-08-27
fix(modal): remove redundant omit that breaks the preset props type check

lazerg:
#8189 switched Modal's `presetProps` to `keep(props, forwardedPresetPropsKeys)`, and those keys already exclude `onClose`, `onNegativeClick` and `onPositiveClick`. That left the `omit()` call below it both redundant and impossible to type, so `tsc` now reports three TS2322 errors on Modal.tsx. The lint and test jobs fail on every PR against main because of it, and since `lint:src-type` is the same `tsc -b tsconfig.esm.json` step `build:package` starts with, the release build on main is broken too. Sorry about that, it only surfaces in a full project type check, not in the modal tests I ran.

Dropping the `omit()` call and its now-unused import clears all three. Nothing changes at runtime, the three keys were already absent from the object `keep()` builds.

One error is left after this: `src/pagination/tests/Pagination.virtual-jump.browser.spec.tsx(53,10): error TS2874`, from an unrelated commit. It still fails the test and test-browser jobs, and I haven't touched it here.

### DIFF
--- a/src/modal/src/Modal.tsx
+++ b/src/modal/src/Modal.tsx
@@ -25,7 +25,6 @@ import {
   call,
   eventEffectNotPerformed,
   keep,
-  omit,
   useIsComposing,
   warnOnce
 } from '../../_utils'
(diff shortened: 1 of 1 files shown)


lazerg:

PR #8206, opened 2026-08-27
test(pagination): import h in the virtual jump browser spec

lazerg:
`Pagination.virtual-jump.browser.spec.tsx` uses JSX but doesn't import `h`, and the repo's tsconfig sets `jsxFactory: "h"`. So `tsc -p tsconfig.test.json --noEmit` fails with `error TS2874: This JSX tag requires 'h' to be in scope, but it could not be found`.

Both the `test` and `test:browser` scripts run that type check before vitest, so the two CI jobs currently fail on every open PR. Adding `h` to the existing vue import fixes it, same as the data-table browser spec next door.


lazerg:
The remaining CI failures are the three `Modal.tsx` TS2322 errors. They predate this change and hit every open PR right now. I've put a fix up in #8205.

lazerg:

